PR #5, opened 2020-02-01
fix compilation and plugin bundle

elfenheart:


CuriousVini (approved):

CuriousVini:
Just to add context the compilation was failing because of corresponding api changes in `createTableAssessor` method. Also the plugin classes were not exposed which caused plugins not being available.

PR #7, opened 2020-02-06
update to use new APIs

albertshau:


CuriousVini (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -35,9 +39,25 @@
...
+      try {
+        String bqType = toBigQueryType(field);
+        columnAssessments.add(ColumnAssessment.builder(field.getName(), bqType)
+                                .setSourceColumn(field.getName())
+                                .build());
+      } catch (Exception e) {
+        columnAssessments.add(ColumnAssessment.builder(field.getName(), "N/A")
+                                .setSourceColumn(field.getName())
+                                .setSupport(ColumnSupport.NO)
+                                .setSuggestion(new ColumnSuggestion(e.getMessage(), Collections.emptyList()))
```
CuriousVini, line 51: should we log the exception as well?

albertshau: in this particular case, no, because it will always be an exception about some unsupported type, and not some connection issue or anything where the root cause is hidden in the stacktrace. I will change it to catch IllegalArgumentException so that it's a little bit more clear

PR #9, opened 2020-02-14
CDAP-16301 handle failures better

albertshau:
Added better error handling. The plugin can fail in several ways:

1. DDL failure. This will throw the exception as-is and let the
app handle retries, up to the pipeline limit.

2. Offset commit failure. The plugin will retry up to the pipeline
limit, then throw a DeltaFailureException, which lets the app know
not to retry and transition to failed state.

3. Failure to write changes to GCS. The failure will be thrown
as-is, letting the pipeline rewind state to the last offset and
retry. This allows the plugin to stream changes directly to GCS
without buffering them in memory.

4. Failure to load GCS changes into staging table. The job will
be retried up to the pipeline limit, then throw a
DeltaFailureException, which lets the app know not to retry and
transition to a failed state.

5. Failure to merge changes from staging table into target table.
The job will be retried up to the pipeline limit, then throw a
DeltaFailureException.

Also added a unit test.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/TableBlob.java
+++ b/src/main/java/io/cdap/delta/bigquery/TableBlob.java
@@ -27,10 +27,10 @@ public class TableBlob {
   private final String table;
   private final Schema stagingSchema;
   private final Schema targetSchema;
-  private final String batchId;
+  private final long batchId;
   private final Blob blob;
 
-  public TableBlob(String dataset, String table, Schema targetSchema, Schema stagingSchema, String batchId, Blob blob) {
+  public TableBlob(String dataset, String table, Schema targetSchema, Schema stagingSchema, long batchId, Blob blob) {
     this.dataset = dataset;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/delta/bigquery/TableBlob.java
```diff
@@ -19,6 +19,8 @@
 import com.google.cloud.storage.Blob;
 import io.cdap.cdap.api.data.schema.Schema;
 
+import java.util.List;
```
shifuxu0301: nit, is this an redundant imports?

shifuxu0301 (approved):
LGTM, just small comment.

src/main/java/io/cdap/delta/bigquery/TableBlob.java
albertshau: yep, removed

PR #10, opened 2020-02-18
CDAP-16305 fix primary key updates

albertshau:
Fixed the algorithm to properly handle updates to primary key
fields. This requires expanding the staging table so that it
includes the previous value of the row as well as the new value.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -19,6 +19,8 @@ package io.cdap.delta.bigquery;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.DatasetId;
 import com.google.cloud.bigquery.DatasetInfo;
+import com.google.cloud.bigquery.FieldValue;
+import com.google.cloud.bigquery.FieldValueList;
 import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobId;
@@ -30,6 +32,7 @@ import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableDefinition;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableInfo;
+import com.google.cloud.bigquery.TableResult;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
@@ -37,7 +40,6 @@ import com.google.common.annotations.VisibleForTesting;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.delta.api.DDLEvent;
 import io.cdap.delta.api.DMLEvent;
-import io.cdap.delta.api.DMLOperation;
 import io.cdap.delta.api.DeltaFailureException;
 import io.cdap.delta.api.DeltaTargetContext;
 import io.cdap.delta.api.EventConsumer;
@@ -58,7 +60,11 @@ import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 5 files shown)


src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,31 @@
...
+  ~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  ~ License for the specific language governing permissions and limitations under
+  ~ the License.
+  -->
+
+<configuration>
+
+  <appender name="Console" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>%d{ISO8601} - %-5p [%t:%C{1}@%L] - %m%n</pattern>
```
shifuxu0301, line 23: Just curious and take this as an education. What is the the "logback" for?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -161,6 +154,8 @@
   private final Bucket bucket;
   private final String project;
   private final RetryPolicy<Object> commitRetryPolicy;
+  private final Map<TableId, Long> latestSeenSequence;
+  private final Map<TableId, Long> latestMergedSequence;
```
shifuxu0301, line 158: Quick question? are these maps thread-safe?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -431,77 +451,140 @@ private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemp
...
+     * | 1234567890     | 6             | INSERT |            |              | 0  | Alice
+     * | 1234567890     | 7             | INSERT |            |              | 1  | blob
+     * | 1234567890     | 8             | UPDATE | 1          | blob         | 1  | Bob
+     *
+     * If the primary key is the 'id' field, the merge is performed by running the following query:
+     *
+     * MERGE [target table] as T
+     * USING ($DIFF_QUERY) as D
+     * ON T.id = D._before_id
+     * WHEN MATCHED AND D._op = "DELETE
```
shifuxu0301: "DELETE"

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -551,4 +635,52 @@ private void runWithRetries(ContextualRunnable runnable, TableBlob blob, String
       throw e;
     }
   }
+
+  private long getLatestSequenceNum(TableId tableId) throws InterruptedException, DeltaFailureException {
+    RetryPolicy<Object> retryPolicy = new RetryPolicy<>()
+      .withMaxDuration(Duration.of(context.getMaxRetrySeconds(), ChronoUnit.SECONDS))
+      // maxDuration must be greater than the delay, otherwise Failsafe complains
+      .withBackoff(Math.min(61, context.getMaxRetrySeconds()) - 1, 120, ChronoUnit.SECONDS)
```
shifuxu0301, line 644: It looks like when maxRetrySecs are less than 60s, we will end up with retry just once? Should we consider retrying more times? maybe at least 2 or 3?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -551,4 +635,52 @@ private void runWithRetries(ContextualRunnable runnable, TableBlob blob, String
...
+    try {
+      return Failsafe.with(retryPolicy).get(() -> {
+        if (bigQuery.getTable(tableId) == null) {
+          return 0L;
+        }
+
+        String query = String.format("SELECT MAX(_sequence_num) FROM %s.%s", tableId.getDataset(), tableId.getTable());
+        QueryJobConfiguration jobConfig = QueryJobConfiguration.of(query);
+        JobId jobId = JobId.of(UUID.randomUUID().toString());
+        Job queryJob = bigQuery.create(JobInfo.newBuilder(jobConfig).setJobId(jobId).build());
```
shifuxu0301, line 660: Just curious, we are creating a big query job with a random UUID as jobId here? Wandering why it is required to have jobId here?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -551,4 +635,52 @@ private void runWithRetries(ContextualRunnable runnable, TableBlob blob, String
...
+        queryJob.waitFor();
+        TableResult result = queryJob.getQueryResults();
+        Iterator<FieldValueList> resultIter = result.iterateAll().iterator();
+        if (!resultIter.hasNext()) {
+          return 0L;
+        }
+        FieldValue val = resultIter.next().get(0);
+        if (val.getValue() == null) {
+          return 0L;
+        }
```
shifuxu0301, line 672: Can we add a comment here to explain what is get(0) for? I think that is for the getting the sequence number.

src/test/resources/logback-test.xml
albertshau: this configures how log messages should be printed in unit tests. Before adding this and logback-classic as a test dependency in the pom, no log messages would show up when running the test.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: they are not, but the methods in this class that access them are synchronized. I can add a comment

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: In this case we are executing bigquery operations, which means the most likely cause of failure is some sort of misconfiguration or permissions issue, which are unlikely to be resolved without manual intervention. So if the user has set a low retry timeout, I think it's fine if we don't do many attempts.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: This is from the example in the docs at https://cloud.google.com/bigquery/docs/quickstarts/quickstart-client-libraries#running_the_query. I haven't tried without setting the ID to see if BQ generates one for you.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: agree. makes sense for me

src/test/resources/logback-test.xml
shifuxu0301: Thanks.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Cool.

shifuxu0301 (approved):
LGTM, Just need to add additional comments and add the quote before merge.

PR #14, opened 2020-03-11
Flush out docs for plugin

albertshau:


widgets/bigquery-cdcTarget.json
```diff
@@ -25,7 +20,17 @@
       ]
     },
     {
-      "label" : "CDC Configuration",
+      "label": "Credentials",
+      "properties": [
+        {
+          "name": "serviceAccountKey",
+          "label": "Service Account Key",
+          "widget-type": "password"
```
shifuxu0301, line 31: The previous widget-type is "textarea", now here is "password". Should we keep using "textarea"?

docs/bigquery-cdcTarget.md
```diff
@@ -16,3 +22,24 @@ The service account key can be found on the Dashboard in the Cloud Platform Cons
 Make sure the account key has permission to access BigQuery.
 The service account key file needs to be available on every node in your cluster and
 must be readable by all users running the job.
+
+Properties
+----------
+
+**Project**: Project of the BigQuery dataset. When running on a Dataproc cluster, this can be left blank,
```
shifuxu0301, line 29: A small question here, "this can be left blank". When I opened some google plugin in hub, it looks like the default value is "auto-detect", should we use that instead?

docs/bigquery-cdcTarget.md
```diff
@@ -16,3 +22,24 @@ The service account key can be found on the Dashboard in the Cloud Platform Cons
...
+
+**Staging Bucket**: GCS bucket to write change events to before loading them into staging tables.
+This bucket can be shared across multiple replicators. The bucket must be in the same location as the
+BigQuery dataset.
+
+**Service Account Key**: The contents of the service account key to use when interacting with GCS and
+BigQuery. When running on a Dataproc cluster, this can be left blank, which will use the service account
+of the cluster.
+
+**Max Batch Seconds**: Maximum number of seconds to wait before merging a batch of changes.
```
shifuxu0301, line 40: Should we also mention default number here?

docs/bigquery-cdcTarget.md
```diff
@@ -16,3 +22,24 @@ The service account key can be found on the Dashboard in the Cloud Platform Cons
...
+
+**Max Batch Changes**: Maximum number of change events to include in a single batch.
+
+**Staging Table Prefix**: Changes are first written to a staging table before merged to the final table.
+Staging tables are named used this prefix prepended to the target table name.
```
shifuxu0301, line 45: Same thing, should we mention default value here?

widgets/bigquery-cdcTarget.json
albertshau: this is supposed to be a secret though, since it is the account key. In the cloud, we expect this to be left blank. In on-prem, we would normally expect this to be a secure macro.

docs/bigquery-cdcTarget.md
albertshau: we no longer mention defaults in the docs, and instead set defaults directly in the widget.

shifuxu0301 (approved):
LGTM

PR #16, opened 2020-03-27
CDAP-16300 Improve primary key handling for BQ target.

shifuxu0301:

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -156,6 +156,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   private final RetryPolicy<Object> commitRetryPolicy;
   private final Map<TableId, Long> latestSeenSequence;
   private final Map<TableId, Long> latestMergedSequence;
+  private final Map<TableId, List<String>> primaryKeyStore;
   private ScheduledExecutorService executorService;
   private ScheduledFuture<?> scheduledFlush;
   private Offset latestOffset;
@@ -185,6 +186,7 @@ public class BigQueryEventConsumer implements EventConsumer {
     // these maps are only accessed in synchronized methods so they do not need to be thread safe.
     this.latestMergedSequence = new HashMap<>();
     this.latestSeenSequence = new HashMap<>();
+    this.primaryKeyStore = new HashMap<>();
     this.commitRetryPolicy = new RetryPolicy<>()
       .withMaxAttempts(Integer.MAX_VALUE)
       .withMaxDuration(Duration.of(5, ChronoUnit.MINUTES))
@@ -241,6 +243,7 @@ public class BigQueryEventConsumer implements EventConsumer {
         break;
       case DROP_DATABASE:
         datasetId = DatasetId.of(project, event.getDatabase());
+        primaryKeyStore.clear();
         if (bigQuery.getDataset(datasetId) != null) {
           bigQuery.delete(datasetId);
         }
@@ -253,14 +256,17 @@ public class BigQueryEventConsumer implements EventConsumer {
           TableDefinition tableDefinition = StandardTableDefinition.newBuilder()
             .setSchema(Schemas.convert(addSequenceNumber(event.getSchema())))
             .build();
-          String primaryKey;
-          if (event.getPrimaryKey().isEmpty()) {
-            primaryKey = "_sequence_num";
-          } else {
-            primaryKey = event.getPrimaryKey().stream().collect(Collectors.joining(","));
+          List<String> primaryKeys = event.getPrimaryKey();
+          if (primaryKeys.isEmpty()) {
+            throw new DeltaFailureException(
+              String.format("Primary key should not be empty for table '%s' in dataset '%s'",
+                            tableId.getTable(), tableId.getDataset()));
           }
+
+          String pkStr = primaryKeys.stream().collect(Collectors.joining(","));
+          primaryKeyStore.put(tableId, primaryKeys);
           context.putState(String.format("%s-%s-primary-key", tableId.getDataset(), tableId.getTable()),
-                           primaryKey.getBytes());
+                           pkStr.getBytes());
           TableInfo tableInfo = TableInfo.newBuilder(tableId, tableDefinition).build();
           bigQuery.create(tableInfo);
         }
@@ -270,6 +276,7 @@ public class BigQueryEventConsumer implements EventConsumer {
         // shouldn't exist
         flush();
         tableId = TableId.of(project, event.getDatabase(), event.getTable());
+        primaryKeyStore.remove(tableId);
         table = bigQuery.getTable(tableId);
         if (table != null) {
           bigQuery.delete(tableId);
@@ -293,7 +300,8 @@ public class BigQueryEventConsumer implements EventConsumer {
         // TODO: alter the staging table as well
         break;
       case RENAME_TABLE:
-        // TODO: flush changes, execute a copy job, delete previous table, drop old staging table
+        // TODO: flush changes, execute a copy job, delete previous table, drop old staging table, remove old entry
+        //  in primaryKeyStore, put new entry in primaryKeyStore
         break;
       case TRUNCATE_TABLE:
         // TODO: flush changes then run a DELETE from table WHERE 1=1 query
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -249,20 +249,19 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
...
-        // TODO: add a way to get PK of a table through Delta API?
         if (table == null) {
           TableDefinition tableDefinition = StandardTableDefinition.newBuilder()
             .setSchema(Schemas.convert(addSequenceNumber(event.getSchema())))
             .build();
-          String description;
+          String primaryKey;
           if (event.getPrimaryKey().isEmpty()) {
-            description = "Primary Key: _sequence_num";
+            primaryKey = "_sequence_num";
```
albertshau: I don't think this actually works, was probably some placeholder I put here in the spike. We should actually throw a DeltaFailureException saying tables without primary keys are not supported.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -249,20 +249,19 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
...
           } else {
-            description = "Primary Key: " + event.getPrimaryKey().stream().collect(Collectors.joining(","));
+            primaryKey = event.getPrimaryKey().stream().collect(Collectors.joining(","));
           }
-          TableInfo tableInfo = TableInfo.newBuilder(tableId, tableDefinition)
-            .setDescription(description)
-            .build();
+          context.putState(String.format("%s-%s-primary-key", tableId.getDataset(), tableId.getTable()),
```
albertshau, line 270: since it's guaranteed that all events for this table will go to the same process, we should keep an in-memory map from tableId to primary key. Here it should write to both the in-memory map and call context.putState().

Later on, when using the primary key, we should first check the in-memory map, and only use context.getState() if it's not there. It's probably worth having some sort of PrimaryKeyStore to abstract this away.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -249,20 +249,19 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
...
             .setSchema(Schemas.convert(addSequenceNumber(event.getSchema())))
             .build();
-          String description;
+          String primaryKey;
           if (event.getPrimaryKey().isEmpty()) {
-            description = "Primary Key: _sequence_num";
+            primaryKey = "_sequence_num";
           } else {
-            description = "Primary Key: " + event.getPrimaryKey().stream().collect(Collectors.joining(","));
+            primaryKey = event.getPrimaryKey().stream().collect(Collectors.joining(","));
```
albertshau: I believe this is the case, but BQ column names can't have commas right?

albertshau (commented):
comment about caching the primary keys in memory

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure, will throw exception.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Yes. BQ column name should not contain comma. I am using the comma here to join multiple primary keys if there is any. When I am get the string of primary key, we will split them by comma. From my testing, it works.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure, I can add in-memory state for primary keys.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: hm actually the primary key here comes from the source. None of our current sources have commas, but there is no guarantee for the future.

I think it's more future-proof to use gson to serialize an entire object, even if it currently only contains a single `List<String>` for the primaryKey. This will let us add new information into the state in future versions if needed and it will still be backwards compatible.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -249,20 +252,22 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
...
             .build();
-          String description;
-          if (event.getPrimaryKey().isEmpty()) {
-            description = "Primary Key: _sequence_num";
-          } else {
-            description = "Primary Key: " + event.getPrimaryKey().stream().collect(Collectors.joining(","));
+          List<String> primaryKeys = event.getPrimaryKey();
+          if (primaryKeys.isEmpty()) {
+            throw new DeltaFailureException(
+              String.format("Primary key should not be empty for table '%s' in dataset '%s'",
```
albertshau: nit: Table '%s' in Database '%s' has no primary key. Tables without a primary key are not supported.

albertshau (commented):
I left another comment about storing an object as the state to be more future-proof. It probably wasn't seen because it responded to a conversation marked as resolved. Can you take a look at that?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -458,26 +465,29 @@ private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemp
...
-                        + "deleted by other processes.", blob.getDataset(), blob.getTable()));
-    }
+    List<String> primaryKey = primaryKeyStore.get(targetTableId);
+    if (primaryKey == null) {
+      byte[] pkBytes;
+      try {
+        pkBytes = context.getState(String.format("%s-%s-primary-key", targetTableId.getDataset(),
+                                                 targetTableId.getTable()));
+      } catch (Exception e) {
+        throw new DeltaFailureException(
```
albertshau: DeltaFailureException will cause the replicator to fail immediately, and should only be thrown if there is no way for the replicator to ever make progress in any scenario. 

For this case, it could potentially be a temporary issue, so we can throw the exception as-is, and let the replicator handle retrying.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -458,26 +465,29 @@ private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemp
...
+                        targetTableId.getTable(), targetTableId.getDataset()), e);
+      }
 
-    // TODO: figure app should provide a way to get table info given an offset
-    String pkStr = targetTable.getDescription();
-    pkStr = pkStr == null || !pkStr.startsWith("Primary Key: ") ?
-      "_sequence_num" : pkStr.substring("Primary Key: ".length());
-    List<String> primaryKey = Arrays.stream(pkStr.split(",")).collect(Collectors.toList());
+      if (pkBytes == null) {
+        throw new DeltaFailureException(
```
albertshau, line 478: this should never happen, unless there was some sort of manual deletion of the state. Let's change the message to:

Primary key information for table '%s' in dataset '%s' could not be found. This can only happen if state was manually deleted. Please create a new replicator and start again.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure. Agree, I will do that.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure. Will apply.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure, makes sense. will do.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure. I will update the message body.

shifuxu0301:
Due to file conflicts, close this PR, will sync the branch and send out a new one.

PR #17, opened 2020-03-27
Add more assessment for potential db or table name normalization.

shifuxu0301:
This PR is for adding more potential problems to alert users during the assessment if the db and table name will have problem after normalization.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
@@ -78,7 +78,7 @@ public class BigQueryAssessor implements TableAssessor<StandardizedTableDetail> 
                     String.format("Database '%s' will be normalized to '%s' to meet BigQuery's dataset name " +
                                     "requirements.", dbName, normalizedDBName),
                     "Verify that multiple databases will not be normalized to the same BigQuery dataset name",
-                    "There will be conflicts if multiple databases would normalize to the same dataset name"));
+                    "If multiple databases are normalized to the same name, conflicts can occur"));
     }
     if (stagingTablePrefix.length() + tableName.length() > BigQueryEventConsumer.MAX_LENGTH ||
       !tableName.matches(BigQueryEventConsumer.VALID_NAME_REGEX) ||
@@ -90,7 +90,7 @@ public class BigQueryAssessor implements TableAssessor<StandardizedTableDetail> 
                                   tableName, normalizedTableName, normalizedStagingTableName),
                     "Verify that multiple tables will not be normalized to the same BigQuery table name " +
                       "under the same dataset",
-                    "There will be conflicts if multiple tables would normalize to the same name"));
+                    "If multiple tables are normalized to the same name, conflicts can occur."));
     }
 
     return new TableAssessment(columnAssessments, problems);
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
@@ -60,6 +60,7 @@ public class BigQueryAssessor implements TableAssessor<StandardizedTableDetail> 
     List<Problem> problems = new ArrayList<>();
     String dbName = tableDetail.getDatabase();
     String tableName = tableDetail.getTable();
+    String stagingTableName = stagingTablePrefix + tableName;
     if (tableDetail.getPrimaryKey().isEmpty()) {
       problems.add(
         new Problem("Missing Primary Key",
@@ -71,8 +72,8 @@ public class BigQueryAssessor implements TableAssessor<StandardizedTableDetail> 
 
     String normalizedDBName = BigQueryEventConsumer.normalize(dbName);
     String normalizedTableName = BigQueryEventConsumer.normalize(tableName);
-    String normalizedStagingTableName = BigQueryEventConsumer.normalize(stagingTablePrefix + tableName);
-    if (dbName.length() > BigQueryEventConsumer.MAX_LENGTH || !dbName.matches(BigQueryEventConsumer.VALID_NAME_REGEX)) {
+    String normalizedStagingTableName = BigQueryEventConsumer.normalize(stagingTableName);
+    if (!dbName.equals(normalizedDBName)) {
       problems.add(
         new Problem("Normalizing Database Name",
                     String.format("Database '%s' will be normalized to '%s' to meet BigQuery's dataset name " +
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +62,58 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+      problems.add(new Problem("Missing Primary Key",
+                               String.format("Table '%s' in database '%s' must have a primary key in order to be " +
+                                               "replicated", dbName, tableName),
+                               "Please alter the table to use a primary key, or select a different table",
+                               "Not able to replicate this table in BigQuery side"));
+    }
+
+    if (dbName.length() > MAX_LENGTH) {
+      problems.add(new Problem("Dataset Name Exceeding Max Length",
+                               String.format("Dataset name '%s' exceeds the max length %d defined in BigQuery",
```
albertshau: Database '%s' will be normalized to '%s' to meet BigQuery's dataset name requirements.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +62,58 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+                                             dbName, MAX_LENGTH),
+                               "It's better to alter the database to make the name shorter",
```
albertshau: Verify that multiple databases will not be shortened to the same BigQuery dataset name.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +62,58 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+                               "There could be issues if multiple datasets would normalize to the same thing " +
```
albertshau: If multiple databases are normalized to the same name, conflicts can occur.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +62,58 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+                                 "which will cause conflicts"));
+    }
+    if (!dbName.matches(VALID_NAME_REGEX)) {
```
albertshau: in most circumstances, users cannot rename their source databases and tables. Instead, the BQ target should normalize them automatically to something that BQ does support (for example, by substituting all invalid characters into underscore). Problems here will just highlight that they are being renamed, and that conflicts may arise.

So these problems and the name length can be combined into a single problem -- one that is really more of a warning but not necessarily an error.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +62,58 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+      problems.add(new Problem("Dataase Name Not Valid",
+                               String.format("Dataset name '%s' is not valid, BigQuery can only contain letters " +
+                                               "(upper or lower case), numbers and underscores", dbName),
+                               "It's better to alter the database to make the name qualified and unique " +
+                                 "across the project",
+                               "There could be issues if multiple datasets would normalize to the same thing " +
+                                 "which will cause conflicts"));
     }
+    if (stagingTablePrefix.length() + tableName.length() > MAX_LENGTH) {
```
albertshau: similar comments for tables

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: Got it. Makes sense for me. I will combine them together into single problem.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: Sure, will do.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: Will apply.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: Sure. Will apply.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +58,41 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+                    String.format("Table '%s' in database '%s' must have a primary key in order to be replicated",
+                                  dbName, tableName),
+                    "Please alter the table to use a primary key, or select a different table",
+                    "Not able to replicate this table in BigQuery side"));
+    }
+
+    String normalizedDBName = BigQueryEventConsumer.normalize(dbName);
+    String normalizedTableName = BigQueryEventConsumer.normalize(tableName);
+    String normalizedStagingTableName = BigQueryEventConsumer.normalize(stagingTablePrefix + tableName);
+    if (dbName.length() > BigQueryEventConsumer.MAX_LENGTH || !dbName.matches(BigQueryEventConsumer.VALID_NAME_REGEX)) {
```
albertshau: here it's simpler and safer (in case normalize() logic changes) to just check if the dbName is not equal to the normalizedDBName

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -53,10 +58,41 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+      problems.add(
+        new Problem("Normalizing Database Name",
+                    String.format("Database '%s' will be normalized to '%s' to meet BigQuery's dataset name " +
+                                    "requirements.", dbName, normalizedDBName),
+                    "Verify that multiple databases will not be normalized to the same BigQuery dataset name",
+                    "If multiple databases are normalized to the same name, conflicts can occur"));
+    }
+    if (stagingTablePrefix.length() + tableName.length() > BigQueryEventConsumer.MAX_LENGTH ||
```
albertshau: similar here, can just check if the tableName is not equal to the normalizedTableName

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -688,4 +696,17 @@ private long getLatestSequenceNum(TableId tableId) throws InterruptedException,
       throw e;
     }
   }
+
+  public static String normalize(String name) {
```
albertshau, line 700: this should also be used when handling DDL and committing a batch of DML events. Or is that going to be in a follow up PR?

shifuxu0301: Yes. Albert. This is in my next PR.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: => "in case normalize() logic changes"
True, this is a very good suggestion, will do that.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: will do.

albertshau (approved):

PR #18, opened 2020-03-31
CDAP-16483 Apply normalized database and table names for BQ delta target.

shifuxu0301:
This pr is for applying normalized db and table name during handling DML and DDL events.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -237,24 +237,25 @@ public class BigQueryEventConsumer implements EventConsumer {
       throw flushException;
     }
 
-    // TODO: sanitize dataset and table names
     DDLEvent event = sequencedEvent.getEvent();
+    String normalizedDatabaseName = normalize(event.getDatabase());
+    String normalizedTableName = normalize(event.getTable());
     switch (event.getOperation()) {
       case CREATE_DATABASE:
-        DatasetId datasetId = DatasetId.of(project, event.getDatabase());
+        DatasetId datasetId = DatasetId.of(project, normalizedDatabaseName);
         if (bigQuery.getDataset(datasetId) == null) {
           DatasetInfo datasetInfo = DatasetInfo.newBuilder(datasetId).setLocation(bucket.getLocation()).build();
           bigQuery.create(datasetInfo);
         }
         break;
       case DROP_DATABASE:
-        datasetId = DatasetId.of(project, event.getDatabase());
+        datasetId = DatasetId.of(project, normalizedDatabaseName);
         if (bigQuery.getDataset(datasetId) != null) {
           bigQuery.delete(datasetId);
         }
         break;
       case CREATE_TABLE:
-        TableId tableId = TableId.of(project, event.getDatabase(), event.getTable());
+        TableId tableId = TableId.of(project, normalizedDatabaseName, normalizedTableName);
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -349,27 +350,41 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
     }
 
     DMLEvent event = sequencedEvent.getEvent();
-    gcsWriter.write(sequencedEvent);
+    String normalizedDatabaseName = normalize(event.getDatabase());
+    String normalizedTableName = normalize(event.getTable());
+    DMLEvent normalizedDMLEvent = DMLEvent.builder()
+      .setOffset(event.getOffset())
```
albertshau: it will be easy to miss something if we add a new property to DMLEvent. Let's add a builder() method that takes an existing DMLEvent and sets up all the Builder fields already. This can then be changed to:

```
DMLEvent normalizedEvent = DMLEvent.builder(event)
  .setDatabase(normalizedDatabaseName)
  .setTable(normalizedTableName)
  .build()
```

That way, if we add any fields in the future, we just need to update that builder() method, we don't need to change any plugins that happen to do something like this.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -406,7 +421,8 @@ synchronized void flush() throws InterruptedException, IOException, DeltaFailure
   }
 
   private void mergeTableChanges(TableBlob blob) throws DeltaFailureException, InterruptedException {
-    TableId stagingTableId = TableId.of(project, blob.getDataset(), stagingTablePrefix + blob.getTable());
+    String normalizedStagingTableName = normalize(stagingTablePrefix + blob.getTable());
```
albertshau, line 417: blob.getTable() should already be normalized right? Is this just to normalize the table prefix? If so, we can normalize it once in the constructor rather than every time here.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: That's a good suggestion. Will do.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: blob.getTable() will be already normalized. There is an edge case that, if the normalized table name has length around 1020 and the normalized staging prefix is 8 for example. Then if we just combine them, it will have 1028 length which will exceed the 1024 limit. That is reason I initially do the normalization here.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: PR out: https://github.com/data-integrations/delta/pull/48

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: I see, makes sense

shifuxu0301: Cool. I have updated the PR with new builder().

albertshau (approved):

PR #19, opened 2020-03-31
CDAP-16300 Improve primary key handling for BQ target.

shifuxu0301:

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -37,6 +37,8 @@ import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.Storage;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.delta.api.DDLEvent;
 import io.cdap.delta.api.DMLEvent;
@@ -146,6 +148,7 @@ import java.util.stream.Collectors;
  */
 public class BigQueryEventConsumer implements EventConsumer {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryEventConsumer.class);
+  private static final Gson GSON = new Gson();
   private static final int MAX_LENGTH = 1024;
   // according to big query dataset and table naming convention, valid name should only contain letters (upper or
   // lower case), numbers, and underscores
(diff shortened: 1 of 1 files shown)


shifuxu0301:
Due to conflicts, I have closed the previous PR (https://github.com/data-integrations/bigquery-delta-plugins/pull/16). send out a new one here.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -257,20 +263,20 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
...
+            throw new DeltaFailureException(
+              String.format("Table '%s' in database '%s' has no primary key. Tables without a primary key are" +
+                              " not supported.", tableId.getTable(), tableId.getDataset()));
           }
-          TableInfo tableInfo = TableInfo.newBuilder(tableId, tableDefinition)
-            .setDescription(description)
-            .build();
+          primaryKeyStore.put(tableId, primaryKeys);
+          context.putState(String.format("%s-%s-primary-key", tableId.getDataset(), tableId.getTable()),
+                           GSON.toJson(primaryKeys).getBytes());
```
albertshau, line 279: what I meant here was a class that has primaryKeys as a field. That way, if we need to add more information, we can just add it to the class and things will be compatible.

shifuxu0301: I will add a class here. 
I have one question related to the state file. Should we store the state per table (which means each big query table will have a state file), or we should store the per table primary key map into single state file?
They are some tradeoffs between these two.  
Method 1: Pros: better isolation, no worry to handle the override issue. Cons: If there are a lot of replicating tables, there will be a lot of files stored on disk.
Method 2: Pros: Single file contains all the primary key information for each table. Cons: every time, we put/update in-memory store, we will read from the local store first and merge the updates about the cache to avoid override.

albertshau: One other thing we need to consider is that GCS has a rate limit of 1 update per second for the same object. So if we have a single file, it could potentially hit that rate limit as well.  With a single file, it also becomes less efficient the more tables we are tracking, as the entire state needs to be stored whenever one primary key is added/updated.

I think it's better to keep them separate.

shifuxu0301: true. agreed. Will keep them separate.

shifuxu0301:
due to conflicts, resend out a new one: https://github.com/data-integrations/bigquery-delta-plugins/pull/20

PR #20, opened 2020-04-01
Feature/cdap 16300 improve the primary key handling for big query target

shifuxu0301:

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -163,6 +163,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   private final RetryPolicy<Object> commitRetryPolicy;
   private final Map<TableId, Long> latestSeenSequence;
   private final Map<TableId, Long> latestMergedSequence;
+  private final Map<TableId, List<String>> primaryKeyStore;
   private ScheduledExecutorService executorService;
   private ScheduledFuture<?> scheduledFlush;
   private Offset latestOffset;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -475,27 +482,22 @@ private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemp
...
-      // this is because the primary key for the table is contained in the DDL event, but not in DML events.
-      // so the primary key info is stored in the table definition.
-      // If we had the primary key information here, we could create the table now if it didn't exist for some reason
-      throw new IllegalStateException(
-        String.format("No target table %s.%s found. Please ensure that tables created by the replicator are not "
-                        + "deleted by other processes.", blob.getDataset(), blob.getTable()));
+    List<String> primaryKey = primaryKeyStore.get(targetTableId);
+    if (primaryKey == null) {
+      byte[] stateBytes = context.getState(String.format("bigquery-%s-%s", targetTableId.getDataset(),
+                                                      targetTableId.getTable()));
```
albertshau: nit: indentation

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -475,27 +482,22 @@ private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemp
...
+      if (stateBytes == null) {
+        throw new DeltaFailureException(
+          String.format("Primary key information for table '%s' in dataset '%s' could not be found. This can only " +
+                          "happen if state was manually deleted. Please create a new replicator and start again.",
```
albertshau: nit: state was manually deleted -> state was corrupted

albertshau (approved):
minor comments, otherwise lgtm

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure, will fix it before merge.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Will update before merging.

PR #24, opened 2020-04-08
Clean staging table after merging is done.

shifuxu0301:
This PR is used to clean up staging table once the merging is done. By doing this, there will be no staging table exists under the big query dataset once the merging is done. 

Tested on big query side, events was able to propagate and apply as expected.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -516,6 +516,9 @@ public class BigQueryEventConsumer implements EventConsumer {
       LOG.warn("Failed to delete temporary GCS object {} in bucket {}. The object will need to be manually deleted.",
                blob.getBlob().getBlobId().getName(), blob.getBlob().getBlobId().getBucket(), e);
     }
+    // clean up staging table after merging is done, there is no retry for this clean up since it will not affect
+    // future functionality
+    bigQuery.delete(stagingTableId);
   }
 
   private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemptNumber) throws InterruptedException {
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -466,6 +466,9 @@ private void mergeTableChanges(TableBlob blob) throws DeltaFailureException, Int
       LOG.warn("Failed to delete temporary GCS object {} in bucket {}. The object will need to be manually deleted.",
                blob.getBlob().getBlobId().getName(), blob.getBlob().getBlobId().getBucket(), e);
     }
+    // clean up staging table after merging is done, there is no retry for this clean up since it will not affect
+    // future functionality
+    bigQuery.delete(stagingTableId);
```
albertshau, line 521: Though I believe this is safe today, I think it would be better if we partition the staging table on the batchid (which is a timestamp), then delete just the data for that partition. It would potentially allow us to build up more data in error scenarios and makes it harder to cause problems if we refactor things in the future. Deleting a partition should be efficient as well.

shifuxu0301: Correct me if I am understanding wrong, do u suggest we will write a query to delete staging  records based on the _batch_id?  In this case, should we also apply runWithRetries with this "delete" query?

albertshau: yeah, though the part that creates the staging table will also have to be changed to specify that it is partitioned on _batch_id. 

This is assuming deleting all data for a specific partition is efficient. It seems like it should be, but I have not tried it myself. Could you try it out with a normal table and a partitioned table to verify? If it's not efficient let's just stick with this delete table logic.

albertshau:
can you also enhance the unit test to check that the staging table is empty/deleted?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: from the other PR, it looks like it's better to just drop the table. Though can you check if there are any quotas for how often you can drop/create tables? If it's not a problem let's just stick this the drop approach.

shifuxu0301:
Close the PR due to the conflicts and reopen a new one.

PR #25, opened 2020-04-09
Implement the feature of TRAUNCATE TABLE in BQ side.

shifuxu0301:
Implement the feature of TRAUNCATE TABLE in BQ side. We do not need to worry about staging table here after the truncate, because we will clear the staging table records in batch every time merging is done.

Tested on local with mysql, it works as expected on big query side, table has been truncated.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -334,7 +334,22 @@ public class BigQueryEventConsumer implements EventConsumer {
         //  in primaryKeyStore, put new entry in primaryKeyStore
         break;
       case TRUNCATE_TABLE:
-        // TODO: flush changes then run a DELETE from table WHERE 1=1 query
+        flush();
+        String query = "DELETE from " + normalizedDatabaseName + "." + normalizedTableName + " WHERE 1=1";
+        QueryJobConfiguration.Builder jobConfigBuilder = QueryJobConfiguration.newBuilder(query);
+        if (encryptionConfig != null) {
+          jobConfigBuilder.setDestinationEncryptionConfiguration(encryptionConfig);
+        }
+        QueryJobConfiguration truncateJobConf = jobConfigBuilder.build();
+        JobId jobId = JobId.newBuilder()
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -334,7 +334,22 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
         //  in primaryKeyStore, put new entry in primaryKeyStore
         break;
       case TRUNCATE_TABLE:
-        // TODO: flush changes then run a DELETE from table WHERE 1=1 query
+        flush();
+        String query = "DELETE from " + normalizedDatabaseName + "." + normalizedTableName + " WHERE 1=1";
```
albertshau, line 338: do you know if this is an efficient query? If it is not, I wonder if deleting the table and re-creating it would be more efficient.

albertshau: can you also add a test case for truncate? Let me know if you have issues running the unit tests.

shifuxu0301: This is a very good callout. 
https://stackoverflow.com/questions/53625705/do-bigquery-charge-if-table-gets-deleted-via-retention-period
By checking this stackoverflow, this kind of delete query will scan the table since the where condition is always true. This is not efficient as delete and re-create table. Also, one thing cause my attention, bigquery is charing for this delete dml query, but is free for delete/create tables. I think we shall delete and recreate the table as a better practice for customer.

shifuxu0301: Yep. I will try to add a test case for truncate. Will let u if I have problem running the tests.

albertshau: Hm I see, if there is a charge for a delete query but not for deleting a table, probably better to ignore my comment about partitioning the staging table in the other PR and just delete it.

shifuxu0301:
Due to conflicts, close this PR and resend a new one.

PR #28, opened 2020-04-10
CDAP-16558 removed max batch size

albertshau:
Since there is a 1000 load job per table per day limit, changed
the plugin to remove the max batch size property. This is no longer
needed anyway, since the plugin streams all its data directly
to GCS and doesn't keep much information in memory.

Added more detailed documentation, and added a feature problem
to the assessment if the load interval is likely too short.

### DIFF
--- a/docs/bigquery-cdcTarget.md
+++ b/docs/bigquery-cdcTarget.md
@@ -11,6 +11,18 @@ The final target tables will include all the original columns from the source ta
 _sequence_num column. The sequence number is used to ensure that data is not duplicated or missed in
 replicator failure scenarios.
 
+Quotas
+------
+BigQuery has a quota of 1,000 operations per table per day, and 100,000 operations per project per day.
+Every load interval, for each target table, this plugin will perform one operation.
+With the default value of 90 seconds, this results in 960 merge jobs for each target table.
+This means if there are no other processes loading data into BigQuery for your project,
+you are limited to replicating 104 tables per day.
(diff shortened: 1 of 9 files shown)


docs/bigquery-cdcTarget.md
```diff
@@ -11,6 +11,18 @@ The final target tables will include all the original columns from the source ta
 _sequence_num column. The sequence number is used to ensure that data is not duplicated or missed in
 replicator failure scenarios.
 
+Quotas
+------
+BigQuery has a quota of 1,000 load jobs per table per day, and 100,000 load jobs per project per day.
+Every load interval, for each target table, this plugin will perform one load job into a staging table,
+followed by a merge job into the target table. With the default value of 90 seconds, 
+this results in 960 staging jobs and 960 merge jobs for each target table. This means if there are no
```
shifuxu0301: 1. since the quota will be reset after every interval we cleaning up the staging table, should we remove the content about the load job limitation on staging table job?

2. Since our current bigquery can not support Alert table operation very well. should we also have a section in docs to say that we are not going to support alter table for now. Docs refer: https://cloud.google.com/bigquery/docs/managing-table-schemas

shifuxu0301 (commented):
nit comments about the docs, the code part LGTM.

docs/bigquery-cdcTarget.md
albertshau: good points, will update

PR #31, opened 2020-04-10
Fix the max delay in RetryPolicy.

shifuxu0301:
Fix the max delay in RetryPolicy. The maxDelay should always be greater than baseDelay. Current code will fail with default loadInterval as 90s.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -826,7 +826,7 @@ public class BigQueryEventConsumer implements EventConsumer {
     }
 
     long baseDelay = Math.min(91, context.getMaxRetrySeconds()) - 1;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -826,7 +826,7 @@ public static String normalize(String name) {
     }
 
     long baseDelay = Math.min(91, context.getMaxRetrySeconds()) - 1;
-    long maxDelay = Math.max(baseDelay, loadIntervalSeconds);
+    long maxDelay = Math.max(baseDelay, 180);
```
albertshau: thanks for fixing. I think this should be Math.max(baseDelay + 1, loadIntervalSeconds)

shifuxu0301: sure, will apply.

albertshau (approved):

PR #33, opened 2020-04-13
CDAP-16570 Update BigQuery Missing Primary Key assessment impact body.

shifuxu0301:


src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -85,7 +85,7 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
                     String.format("Table '%s' in database '%s' must have a primary key in order to be replicated",
                                   tableName, dbName),
                     "Please alter the table to use a primary key, or select a different table",
-                    "Not able to replicate this table in BigQuery side"));
+                    "Not able to replicate this table in BigQuery"));
```
albertshau: nit: I think it's also better to say 'to BigQuery' instead of 'in BigQuery'

albertshau (approved):
minor comment, lgtm

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
shifuxu0301: sure. will update before merging.

PR #36, opened 2020-04-15
CDAP-16575 add an option to require manual intervention for drops

albertshau:
Added an option that will cause the plugin to fail and retry
whenever it sees a drop table or drop database event and the
table/dataset exists. This means dropping those resources will
require manual intervention, but may be desirable for users that
with to ensure that data is not accidentally deleted.


docs/bigquery-cdcTarget.md
```diff
@@ -49,21 +49,29 @@ on the new primary key.
...
 ----------
 
-**Project**: Project of the BigQuery dataset. When running on a Dataproc cluster, this can be left blank,
+**Project ID**: Project of the BigQuery dataset. When running on a Dataproc cluster, this can be left blank,
 which will use the project of the cluster.
 
 **Staging Bucket**: GCS bucket to write change events to before loading them into staging tables.
-This bucket can be shared across multiple replicators. The bucket must be in the same location as the
-BigQuery dataset.
+Changes are written to a directory that contains the replicator name and namespace. It is safe use
```
shifuxu0301: nit: "It is safe to use"

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -268,6 +270,13 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
         datasetId = DatasetId.of(project, normalizedDatabaseName);
         primaryKeyStore.clear();
         if (bigQuery.getDataset(datasetId) != null) {
+          if (requireManualDrops) {
+            throw new IllegalArgumentException(
```
shifuxu0301: Nit, should we throw general RuntimeException rather than IllegalArgumentException, seems like the case does not like an illegal argument case.

shifuxu0301 (approved):
Overall LGTM, small comments.

PR #39, opened 2020-04-29
Make write object batch into GCS in parallel.

shifuxu0301:
This is an improvement that we will reuse the same cached thread pool in GCS writer to write batches in parallel. 

Testing: unit test passed.

Note: I also tested using fixed thread pool in my local, did not see any improvements there, so in this PR to make sure we can release on time, i will reuse the same cached thread pool. I will do further investigation and testing against CDF as next step after this PR.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
+++ b/src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
@@ -23,7 +23,6 @@ import com.google.cloud.storage.Storage;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.delta.api.DMLEvent;
-import io.cdap.delta.api.DMLOperation;
 import io.cdap.delta.api.DeltaTargetContext;
 import io.cdap.delta.api.ReplicationError;
 import io.cdap.delta.api.Sequenced;
@@ -37,10 +36,15 @@ import java.io.OutputStream;
 import java.nio.channels.Channels;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 
 /**
  * Writes to multiple GCS files.
@@ -53,15 +57,17 @@ public class MultiGCSWriter {
   private final Map<Key, TableObject> objects;
   private final Map<Schema, org.apache.avro.Schema> schemaMap;
   private final DeltaTargetContext context;
+  private final ExecutorService executorService;
 
   public MultiGCSWriter(Storage storage, String bucket, String baseObjectName,
-                        DeltaTargetContext context) {
+                        DeltaTargetContext context, ExecutorService executorService) {
     this.storage = storage;
     this.bucket = bucket;
     this.baseObjectName = baseObjectName;
     this.objects = new HashMap<>();
     this.schemaMap = new HashMap<>();
     this.context = context;
+    this.executorService = executorService;
   }
 
   public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -76,35 +82,71 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
...
       TableObject tableObject = entry.getValue();
+      writeFutures.add(executorService.submit((Callable<Void>) () -> {
+        LOG.debug("Writing batch {} of {} events into GCS for table {}.{}", tableObject.batchId, tableObject.numEvents,
+                  tableObject.dataset, tableObject.table);
+        try {
+          tableObject.close();
+        } catch (IOException e) {
+          String errMsg = String.format("Error writing batch of %d changes for %s.%s to GCS",
+                                        tableObject.numEvents, tableObject.dataset, tableObject.table);
+          LOG.error(errMsg, e);
```
chtyim: Don't log if you throw. You can wrap the caught `IOException` with a new message and add the original IOExceptio as the cause.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -76,35 +82,71 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
     }
   }
 
-  public synchronized Collection<TableBlob> flush() throws IOException {
-    List<TableBlob> writtenObjects = new ArrayList<>(objects.size());
-    IOException error = null;
+  public synchronized Collection<TableBlob> flush() throws IOException, InterruptedException {
+    List<TableBlob> writtenObjects = Collections.synchronizedList(new ArrayList<>(objects.size()));
```
chtyim: Instead of having a separate list, you can have the `Callable` return the `TableBlob` object, which you can get through the future.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
shifuxu0301: Sure. will wrap and rethrow.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -76,35 +82,71 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
...
+
+        Blob blob = storage.get(tableObject.blobId);
+        writtenObjects.add(new TableBlob(tableObject.dataset, tableObject.table, tableObject.targetSchema,
+                                         tableObject.stagingSchema, tableObject.batchId, tableObject.numEvents, blob));
+        return null;
+      }));
+    }
+
+    IOException error = null;
+    for (Future writeFuture: writeFutures) {
```
albertshau: style: insert space before ':'

shifuxu0301: sure.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
shifuxu0301: Sure. will do.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -76,35 +80,69 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
     }
   }
 
-  public synchronized Collection<TableBlob> flush() throws IOException {
+  public synchronized Collection<TableBlob> flush() throws IOException, InterruptedException {
     List<TableBlob> writtenObjects = new ArrayList<>(objects.size());
-    IOException error = null;
+    List<Future<?>> writeFutures = new ArrayList<>(objects.size());
```
chtyim: Should be `Future<TableBlob>`

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -76,35 +80,69 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
...
+                  tableObject.dataset, tableObject.table);
+
+        Blob blob = storage.get(tableObject.blobId);
+        return new TableBlob(tableObject.dataset, tableObject.table, tableObject.targetSchema,
+                             tableObject.stagingSchema, tableObject.batchId, tableObject.numEvents, blob);
+      }));
+    }
+
+    IOException error = null;
+    for (Future writeFuture : writeFutures) {
```
chtyim: Use `Future<TableBlob>` as the type

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
shifuxu0301: ops. forgot to update that. will do.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
shifuxu0301: will do here also.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
shifuxu0301: done update the PR.

albertshau (approved):

PR #44, opened 2020-06-25
[CDAP-17016] Use primary keys as clustering columns to improve join perf for BQ target plugin.

shifuxu0301:
This PR is created as an improvement action for BQ target plugin. During merging process, we want to use primary keys as clustering columns to improve our query performance. Note here, since BQ has limitation of having max 4 clustering columns, so for table which has more than 4 primary keys, to achieve best effort, we ill use the first 4 primary key fields as clustering.  In addition, we will pass the max clustering columns as a runtime arguments.

Testing: added unit test passed.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -156,7 +156,6 @@ import javax.annotation.Nullable;
 public class BigQueryEventConsumer implements EventConsumer {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryEventConsumer.class);
   private static final int MAX_LENGTH = 1024;
-  private static final int MAX_CLUSTERING_COLUMNS = 4;
   // according to big query dataset and table naming convention, valid name should only contain letters (upper or
   // lower case), numbers, and underscores
   private static final String VALID_NAME_REGEX = "[\\w]+";
@@ -177,6 +176,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   private final Map<TableId, List<String>> primaryKeyStore;
   private final boolean requireManualDrops;
   private final long baseRetryDelay;
+  private final Integer maxClusteringColumns;
   private ScheduledExecutorService scheduledExecutorService;
   private ScheduledFuture<?> scheduledFlush;
   private ExecutorService executorService;
@@ -220,6 +220,8 @@ public class BigQueryEventConsumer implements EventConsumer {
                                         String.format("cdap/delta/%s/", context.getApplicationName()),
                                         context, executorService);
     this.baseRetryDelay = baseRetryDelay == null ? 10L : baseRetryDelay;
+    String maxClusteringColumnsStr = context.getRuntimeArguments().get("gcp.bigquery.max.clustering.columns");
+    this.maxClusteringColumns =  maxClusteringColumnsStr == null ? null : Integer.valueOf(maxClusteringColumnsStr);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -155,6 +156,7 @@
 public class BigQueryEventConsumer implements EventConsumer {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryEventConsumer.class);
   private static final int MAX_LENGTH = 1024;
+  private static final int MAX_CLUSTERING_COLUMNS = 4;
```
albertshau: can we make this overridable through a runtime argument? I'm thinking it would be better in case BQ increases the max cluster number. then we don't need a code change to take advantage, we could tell users to set a preference, and then restart their replicators.

albertshau (commented):
can you also modify a unit test to check that the table created is clustered on the primary key.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: Sure, this is a good call out. We can definitely do that. will update.

shifuxu0301: Also added unit test in following commit.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -218,6 +220,8 @@
                                         String.format("cdap/delta/%s/", context.getApplicationName()),
                                         context, executorService);
     this.baseRetryDelay = baseRetryDelay == null ? 10L : baseRetryDelay;
+    String maxClusteringColumnsStr = context.getRuntimeArguments().get("gcp.bigquery.max.clustering.columns");
+    this.maxClusteringColumns =  maxClusteringColumnsStr == null ? null : Integer.valueOf(maxClusteringColumnsStr);
```
albertshau: nit: remove extra space before maxClusteringColumnsStr

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -175,6 +176,7 @@
   private final Map<TableId, List<String>> primaryKeyStore;
   private final boolean requireManualDrops;
   private final long baseRetryDelay;
+  private final Integer maxClusteringColumns;
```
albertshau: Integer -> int

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
albertshau: This should default to 4 if nothing is set. If the user explicitly sets it to 0 or less, then we can skip the clustering.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: sure.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
shifuxu0301: I have thought about this, initially I was thinking, we will also need to change the default 4 in the feature if bigquery has bumped up the limit. I took another look of this, I think we can use 4 as default like you said, in this case, like you said, we will always had some baseline improvements.  Will update this default logic!

albertshau (approved):

PR #57, opened 2020-12-01
CDAP-17407 Update bigquery target plugin to handle un-ordered events.

sagarkapare:
This PR is based on the app changes from https://github.com/data-integrations/delta/pull/81. 

Depending on the ordering of events generated by source, staging table will have different schema. 

If the source used in the pipeline is generating `UN-ORDERED` events, then for source table with fields (id, name), staging table would contain following columns:
```
          _op (string)
          _batch_id (long)
          _sequence_num (long)
          _source_timestamp (long)
          _row_id (string)
          id (long)
          name (string)
```

If the source used in the pipeline is generating `ORDERED` events, then for source table with fields (id, name), staging table would contain following columns (this is unchanged):
```
          _op (string)
          _batch_id (long)
          _sequence_num (long)
          id (long)
          name (string)
          _before_id (long)
          _before_name (string)
```

Target table will have following schema irrespective of the ordering of source:

```
_sequence_num (long) 
_is_deleted (boolean  and null for ORDERED source)
_row_id (string  and null for ORDERED source)
_source_timestamp (long and null for ORDERED source)
id (long)
name (string)
```

Sample merge query generated when the source is ORDERED (this query is same as current implementation):
```
MERGE cdc.targettableordered as T
USING (SELECT A.* FROM
(SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as A
LEFT OUTER JOIN
(SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as B
ON A.id = B._before_id AND A.name = B._before_name AND A._sequence_num < B._sequence_num
WHERE B._before_id IS NULL AND B._before_name IS NULL) as D
ON T.id = D._before_id AND T.name = D._before_name
WHEN MATCHED AND D._op = "DELETE" THEN
  DELETE
WHEN MATCHED AND D._op IN ("INSERT", "UPDATE")  THEN
  UPDATE SET _sequence_num = D._sequence_num, id = D.id, name = D.name, id1 = D.id1
WHEN NOT MATCHED AND D._op IN ("INSERT", "UPDATE") THEN
  INSERT (_sequence_num, id, name, id1) VALUES (_sequence_num, id, name, id1)
```

Sample merge query generated when the source is UN_ORDERED:
```
MERGE cdc.targettableordered as T
USING (SELECT A.* FROM
(SELECT * FROM cdc.unorderedstagingtbl WHERE _batch_id = 1111 AND _sequence_num > 10) as A
LEFT OUTER JOIN
(SELECT * FROM cdc.unorderedstagingtbl WHERE _batch_id = 1111 AND _sequence_num > 10) as B
ON A._row_id = B._row_id AND A._source_timestamp < B._source_timestamp 
WHERE  B._row_id IS NULL ) as D
ON  T._row_id = D._row_id 
WHEN MATCHED AND D._op = "DELETE" THEN
  UPDATE SET _is_deleted = true 
WHEN MATCHED AND D._op IN ("INSERT", "UPDATE")  AND D._source_timestamp > T._source_timestamp AND T._is_deleted = false  THEN
  UPDATE SET _sequence_num = D._sequence_num, _source_timestamp = D._source_timestamp, id = D.id, name = D.name, id1 = D.id1
WHEN NOT MATCHED AND D._op IN ("INSERT", "UPDATE") THEN
  INSERT (_sequence_num, _row_id, _source_timestamp, id, name, id1) VALUES (_sequence_num, _row_id, _source_timestamp, id, name, id1)
```

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -51,6 +51,7 @@ import io.cdap.delta.api.EventConsumer;
 import io.cdap.delta.api.Offset;
 import io.cdap.delta.api.ReplicationError;
 import io.cdap.delta.api.Sequenced;
+import io.cdap.delta.api.SourceProperties;
 import net.jodah.failsafe.Failsafe;
 import net.jodah.failsafe.FailsafeException;
 import net.jodah.failsafe.RetryPolicy;
@@ -78,6 +79,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
@@ -178,6 +180,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   private final boolean requireManualDrops;
   private final long baseRetryDelay;
   private final int maxClusteringColumns;
+  private final SourceProperties.Ordering sourceEventOrdering;
   private ScheduledExecutorService scheduledExecutorService;
   private ScheduledFuture<?> scheduledFlush;
   private ExecutorService executorService;
@@ -225,6 +228,8 @@ public class BigQueryEventConsumer implements EventConsumer {
     // current max clustering columns is set as 4 in big query side, use that as default max value
     // https://cloud.google.com/bigquery/docs/creating-clustered-tables#limitations
     this.maxClusteringColumns = maxClusteringColumnsStr == null ? 4 : Integer.parseInt(maxClusteringColumnsStr);
+    this.sourceEventOrdering = context.getSourceProperties() == null ? SourceProperties.Ordering.ORDERED :
+      context.getSourceProperties().getOrdering();
   }
 
   @Override
@@ -339,7 +344,7 @@ public class BigQueryEventConsumer implements EventConsumer {
               .setFields(primaryKeys.subList(0, Math.min(maxClusteringColumns, primaryKeys.size())))
               .build();
           TableDefinition tableDefinition = StandardTableDefinition.newBuilder()
-            .setSchema(Schemas.convert(addSequenceNumber(event.getSchema())))
+            .setSchema(Schemas.convert(addSupplementaryColumnsToTargetSchema(event.getSchema())))
             .setClustering(clustering)
             .build();
 
@@ -388,7 +393,7 @@ public class BigQueryEventConsumer implements EventConsumer {
             .setFields(primaryKeys.subList(0, Math.min(maxClusteringColumns, primaryKeys.size())))
             .build();
         TableDefinition tableDefinition = StandardTableDefinition.newBuilder()
-          .setSchema(Schemas.convert(addSequenceNumber(event.getSchema())))
+          .setSchema(Schemas.convert(addSupplementaryColumnsToTargetSchema(event.getSchema())))
           .setClustering(clustering)
           .build();
         TableInfo.Builder builder = TableInfo.newBuilder(tableId, tableDefinition);
@@ -424,7 +429,7 @@ public class BigQueryEventConsumer implements EventConsumer {
               .setFields(primaryKeys.subList(0, Math.min(maxClusteringColumns, primaryKeys.size())))
               .build();
           tableDefinition = StandardTableDefinition.newBuilder()
-            .setSchema(Schemas.convert(addSequenceNumber(event.getSchema())))
+            .setSchema(Schemas.convert(addSupplementaryColumnsToTargetSchema(event.getSchema())))
             .setClustering(clustering)
             .build();
         }
@@ -471,9 +476,11 @@ public class BigQueryEventConsumer implements EventConsumer {
     return primaryKeys;
   }
 
-  private Schema addSequenceNumber(Schema original) {
-    List<Schema.Field> fields = new ArrayList<>(original.getFields().size() + 1);
-    fields.add(Schema.Field.of("_sequence_num", Schema.of(Schema.Type.LONG)));
+  private Schema addSupplementaryColumnsToTargetSchema(Schema original) {
+    List<Schema.Field> fields = new ArrayList<>(original.getFields().size() + 3);
+    fields.add(Schema.Field.of(Constants.SEQUENCE_NUM, Schema.of(Schema.Type.LONG)));
+    fields.add(Schema.Field.of(Constants.IS_DELETED, Schema.nullableOf(Schema.of(Schema.Type.BOOLEAN))));
+    fields.add(Schema.Field.of(Constants.ROW_ID, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
     fields.addAll(original.getFields());
     return Schema.recordOf(original.getRecordName() + ".sequenced", fields);
   }
@@ -655,6 +662,10 @@ public class BigQueryEventConsumer implements EventConsumer {
     /*
      * Merge data from staging BQ table into target table.
      *
+     * Two independent cases to be considered while performing merge operation:
+     *
+     * Case 1: Source generates ordered events.
+     *
      * If the source table has two columns -- id and name -- the staging table will look something like:
      *
      * | _batch_id      | _sequence_num | _op    | _before_id | _before_name | id | name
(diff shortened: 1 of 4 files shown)


seanzhougoogle:
some questions:

1. do we need the "_source_timestamp (long)"  ? can we just use source_timestamp to generate the sequence_num for unordered events. I think the purpose of them are to identify the order of the events , right ?
2. for ORDERED events, merge query is not that correct:
   a. diff query is wrong:
       ```
         SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as B
         LEFT OUTER JOIN
         ON A.id = B._before_id AND A.name = B._before_name AND A._sequence_num < B._sequence_num
         WHERE B._before_id IS NULL AND B._before_name IS NULL
       ```
       the "_before_*" attribute in the result catch the last value instead of initial value. for example if you have two event
```
       _op              _before_id        _before_name                   id        name                 
       UPDATE                  0                      alice                        1         tom
       UPDATE                  1                       tom                         2         tim
```
     diff query will have below row :
```
           _op              _before_id        _before_name          id        name   
             UPDATE              1                      tom                     2         tim
```
     then it can't be merged to target table. Because in target table , there is  (0, alice) , but (1, tom). 
     so the expected diff query result should be :
```
          _op              _before_id        _before_name              id        name   
             UPDATE              o                      alice                     2         tim
```
    I don't have very good idea so far. looks like we still need some column (like row id) that won't be changed at all. 
    some rough idea: the current diff query actually can get the last change event of a same row , and if we change it to `RIGHT OUTER JOIN` and `WHERE A.id IS NULL AND A.name IS NULL`, we can get the first change event of the same row. but how can we connect the first change event with last change event ? ( we need the _before_id , _before_name from first change event and id , name from last change event of the same row)
    b. nit: theoretically, when MATCHED , INSERT is not possible, because it doesn't have before values (thus won't match any existing row)

3. for Merge query of ORDERED events, do we need  `T._is_deleted = false` if we already have `D._source_timestamp > T._source_timestamp` for MATCHED case ?

sagarkapare:
Thanks Sean for taking look. Have removed the source_timestamp column and the condition `T._is_deleted` based on recommendation in 3rd point above.

New queries looks like below:

For unordered case:
```
MERGE cdc.targettableordered as T
USING (SELECT A.* FROM
(SELECT * FROM cdc.unorderedstagingtbl WHERE _batch_id = 1111 AND _sequence_num > 10) as A
LEFT OUTER JOIN
(SELECT * FROM cdc.unorderedstagingtbl WHERE _batch_id = 1111 AND _sequence_num > 10) as B
ON A._row_id = B._row_id  AND A._sequence_num < B._sequence_num
WHERE  B._row_id IS NULL ) as D
ON  T._row_id = D._row_id 
WHEN MATCHED AND D._op = "DELETE" THEN
  UPDATE SET _is_deleted = true 
WHEN MATCHED AND D._op IN ("INSERT", "UPDATE") AND D._sequence_num > T._sequence_num THEN
  UPDATE SET _sequence_num = D._sequence_num, _row_id = D._row_id, id = D.id, name = D.name, id1 = D.id1
WHEN NOT MATCHED AND D._op IN ("INSERT", "UPDATE") THEN
  INSERT (_sequence_num, _row_id, id, name, id1) VALUES (_sequence_num, _row_id, id, name, id1)
```

For ordered case: 
```
MERGE cdc.targettableordered as T
USING (SELECT A.* FROM
(SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as A
LEFT OUTER JOIN
(SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as B
ON A.id = B._before_id AND A.name = B._before_name AND A._sequence_num < B._sequence_num
WHERE B._before_id IS NULL AND B._before_name IS NULL) as D
ON T.id = D._before_id AND T.name = D._before_name
WHEN MATCHED AND D._op = "DELETE" THEN
  DELETE
WHEN MATCHED AND D._op IN ("INSERT", "UPDATE") AND D._sequence_num > T._sequence_num THEN
  UPDATE SET _sequence_num = D._sequence_num, id = D.id, name = D.name, id1 = D.id1
WHEN NOT MATCHED AND D._op IN ("INSERT", "UPDATE") THEN
  INSERT (_sequence_num, id, name, id1) VALUES (_sequence_num, id, name, id1)
```

Great finding on point 2!!! This will happen if there are multiple updates to primary key in the same batch right? Since its an existing behavior will file separate bug to fix that.

seanzhougoogle:
> Thanks Sean for taking look. Have removed the source_timestamp column and the condition `T._is_deleted` based on recommendation in 3rd point above.
> 
> New queries looks like below:
> 
> For unordered case:
> 
> ```
> MERGE cdc.targettableordered as T
> USING (SELECT A.* FROM
> (SELECT * FROM cdc.unorderedstagingtbl WHERE _batch_id = 1111 AND _sequence_num > 10) as A
> LEFT OUTER JOIN
> (SELECT * FROM cdc.unorderedstagingtbl WHERE _batch_id = 1111 AND _sequence_num > 10) as B
> ON A._row_id = B._row_id  AND A._sequence_num < B._sequence_num
> WHERE  B._row_id IS NULL ) as D
> ON  T._row_id = D._row_id 
> WHEN MATCHED AND D._op = "DELETE" THEN
>   UPDATE SET _is_deleted = true 
> WHEN MATCHED AND D._op IN ("INSERT", "UPDATE") AND D._sequence_num > T._sequence_num THEN
>   UPDATE SET _sequence_num = D._sequence_num, _row_id = D._row_id, id = D.id, name = D.name, id1 = D.id1
> WHEN NOT MATCHED AND D._op IN ("INSERT", "UPDATE") THEN
>   INSERT (_sequence_num, _row_id, id, name, id1) VALUES (_sequence_num, _row_id, id, name, id1)
> ```
> 
> For ordered case:
> 
> ```
> MERGE cdc.targettableordered as T
> USING (SELECT A.* FROM
> (SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as A
> LEFT OUTER JOIN
> (SELECT * FROM cdc.stagingtblordered WHERE _batch_id = 1111 AND _sequence_num > 10) as B
> ON A.id = B._before_id AND A.name = B._before_name AND A._sequence_num < B._sequence_num
> WHERE B._before_id IS NULL AND B._before_name IS NULL) as D
> ON T.id = D._before_id AND T.name = D._before_name
> WHEN MATCHED AND D._op = "DELETE" THEN
>   DELETE
> WHEN MATCHED AND D._op IN ("INSERT", "UPDATE") AND D._sequence_num > T._sequence_num THEN
>   UPDATE SET _sequence_num = D._sequence_num, id = D.id, name = D.name, id1 = D.id1
> WHEN NOT MATCHED AND D._op IN ("INSERT", "UPDATE") THEN
>   INSERT (_sequence_num, id, name, id1) VALUES (_sequence_num, id, name, id1)
> ```
> 
> Great finding on point 2!!! This will happen if there are multiple updates to primary key in the same batch right? Since its an existing behavior will file separate bug to fix that.

Thanks , Sagar, for point 2, yes , and also it will happen if multiple updates of the same row in the same batch if there is no primary key.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -698,69 +709,63 @@ private void mergeStagingTable(TableId stagingTableId, TableBlob blob,
...
-        .stream().map(Schema.Field::getName)
-        .collect(Collectors.joining(", ")) +
-      ") VALUES (" +
-      blob.getTargetSchema().getFields()
-        .stream().map(Schema.Field::getName)
-        .collect(Collectors.joining(", ")) + ")";
+    String diffQuery = createDiffQuery(stagingTableId, primaryKeys, blob.getBatchId(),
+                                       latestMergedSequence.get(targetTableId), sourceEventOrdering);
+
+    String query = createMergeQuery(targetTableId, primaryKeys, blob.getTargetSchema(), diffQuery, sourceEventOrdering);
```
seanzhougoogle: nit: rename query to mergeQuery ?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -781,6 +786,157 @@ private void mergeStagingTable(TableId stagingTableId, TableBlob blob,
...
+       */
+      mergeCondition = " T._row_id = D._row_id ";
+      deleteOperation = "  UPDATE SET _is_deleted = true ";
+    } else {
+      /*
+       * Merge query will be of the following form:
+       *
+       * MERGE [target table] as T
+       * USING ($DIFF_QUERY) as D
+       * ON T.id = D._before_id
```
seanzhougoogle, line 879: T.row_id = D.row_id

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -781,6 +786,157 @@ private void mergeStagingTable(TableId stagingTableId, TableBlob blob,
...
+    if (sourceEventOrdering == SourceProperties.Ordering.UN_ORDERED) {
+      /*
+       * Merge query will be of the following form:
+       *
+       * MERGE [target table] as T
+       *  USING ($DIFF_QUERY) as D
+       * ON T._row_id = D._row_id
+       * WHEN MATCHED AND D.op = “DELETE”
+       *    UPDATE _is_deleted = true
+       * WHEN MATCHED AND D.op IN (“INSERT”, “UPDATE”)
```
seanzhougoogle: Do we need "INSERT" here ?  In which case an INSERT op will match an existing row ?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -781,6 +786,157 @@ private void mergeStagingTable(TableId stagingTableId, TableBlob blob,
...
+      return !field.getName().equals(Constants.ROW_ID);
+    };
+
+    return "MERGE " +
+      targetTableId.getDataset() + "." + targetTableId.getTable() + " as T\n" +
+      "USING (" + diffQuery + ") as D\n" +
+      "ON " + mergeCondition + "\n" +
+      "WHEN MATCHED AND D._op = \"DELETE\" THEN\n" +
+      deleteOperation + "\n" +
+      "WHEN MATCHED AND D._op IN (\"INSERT\", \"UPDATE\") AND D._sequence_num > T._sequence_num THEN\n" +
```
seanzhougoogle: "INSERT" looks not necessary here ?

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -250,16 +259,40 @@ private Schema getTargetSchema(Schema schema) {
...
+      // add all fields from source schema
       fields.addAll(schema.getFields());
+
+      if (sourceEventOrdering == SourceProperties.Ordering.UN_ORDERED) {
+        // add _row_id and _source_timestamp fields to handle un-ordered events
+        fields.add(Schema.Field.of(Constants.ROW_ID, Schema.of(Schema.Type.STRING)));
+        return Schema.recordOf(schema.getRecordName() + ".staging", fields);
+      }
+
+      // add _before_ fields for ORDERED events only
```
seanzhougoogle, line 294: else {
 ...
}

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sagarkapare: Its possible in the scenario where in the same batch if the record with primary key x is deleted and new record with the same primary key is inserted. New record will have _op as INSERT.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sagarkapare: Its possible in the scenario where in the same batch if the record with primary key x is deleted and new record with the same primary key is inserted. New record will have _op as INSERT.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -250,16 +259,40 @@ private Schema getTargetSchema(Schema schema) {
...
+      fields.add(Schema.Field.of(Constants.BATCH_ID, Schema.of(Schema.Type.LONG)));
+      fields.add(Schema.Field.of(Constants.SEQUENCE_NUM, Schema.of(Schema.Type.LONG)));
+
+      // add all fields from source schema
       fields.addAll(schema.getFields());
+
+      if (sourceEventOrdering == SourceProperties.Ordering.UN_ORDERED) {
+        // add _row_id and _source_timestamp fields to handle un-ordered events
+        fields.add(Schema.Field.of(Constants.ROW_ID, Schema.of(Schema.Type.STRING)));
+        return Schema.recordOf(schema.getRecordName() + ".staging", fields);
```
seanzhougoogle: nit: 
How about we do :

if (unordered ) {
   add row_id filed
} else {
   add before fileds
}
return Schema.recordOf(schema.getRecordName() + ".staging", fields);

this will make the code more readable and avoid the duplicate of the "return" statement

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -250,16 +259,40 @@ private Schema getTargetSchema(Schema schema) {
...
+      }
+
+      // add _before_ fields for ORDERED events only
       for (Schema.Field field : schema.getFields()) {
-        if ("_sequence_num".equals(field.getName())) {
-          continue;
-        }
         String beforeName = "_before_" + field.getName();
         Schema beforeSchema = field.getSchema();
         beforeSchema = beforeSchema.isNullable() ? beforeSchema : Schema.nullableOf(beforeSchema);
```
seanzhougoogle: don't quite understand why we need this line ? if it's nullable , that means it's already a UNION type of one type and NULL.

seanzhougoogle (commented):
Thank you Sagar :)

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: in this case , you won't be able to MATCH. for ORDERED case, INSERT event won't have value of "_before" columns. for UNORDERED case, row_id will be different between the newly inserted row and old deleted row.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: the diff query will have two row one with _op=delete and one with _op=insert in this case. only the row with _op = delete has value in "_before" columns and the row with _op=insert won't have any value in "_before" columns. Thus the row  with _op=delete will match the existing row in target table and that row got deleted. row with _op=insert won't match any existing row in target table.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
sagarkapare: This is to make any field in the staging schema of type `_before_<fieldname>` nullable. So for example if `address` field in the input schema is already nullable then we dont need to take any additional changes however say `id` field in the input schema is not nullable we need to make `_before_id` nullable by calling `Schema.nullableOf(beforeSchema)Schema.nullableOf(beforeSchema)`

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
sagarkapare: I agree in general about having single return statements for method for readability, however for cases if we can simplify nesting would be better to  return early. In this case it adds one level more nesting for `else` which is probably ok. updated the code.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sagarkapare: ah. I thought comment is for `NOT MATCHED` where we do need insert and update. yes here we dont. good catch! fixed the code.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
seanzhougoogle: ah, I see, thanks for explanation :)

seanzhougoogle (approved):

seanzhougoogle (approved):

sagarkapare:
https://builds.cask.co/browse/DEL-BPBT56-4
Build is green. Squashing and merging.

PR #69, opened 2020-12-18
Report problem if tables with same name are assessed

seanzhougoogle:


src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -29,13 +28,18 @@
...
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 
 /**
  * Assesses table information.
  */
 public class BigQueryAssessor implements TableAssessor<StandardizedTableDetail> {
   private final String stagingTablePrefix;
+  // tables already assessed so far, key is table
```
sagarkapare: nit: add doc about what is value too or better rename variable name to something like `tableSchema`. Also better initialize in constructor at one place.

Meaning 
private final Map<String, String> tableSchema;

and in constructor do
tableSchema = new HashMap<>();

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -617,6 +618,7 @@ private void loadStagingTable(TableId stagingTableId, TableBlob blob, int attemp
       Clustering clustering = maxClusteringColumns <= 0 ? null : Clustering.newBuilder()
         .setFields(primaryKeys.subList(0, Math.min(maxClusteringColumns, primaryKeys.size())))
         .build();
+      LOG.error("##SEAN create taging table ..." + blob.getStagingSchema().toString());
```
sagarkapare: nit: remove debug logs.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -342,6 +342,7 @@ private void handleDDL(DDLEvent event, String normalizedDatabaseName, String nor
             Clustering.newBuilder()
               .setFields(primaryKeys.subList(0, Math.min(maxClusteringColumns, primaryKeys.size())))
               .build();
+          LOG.error("##SEAN create target table ..." + event.getSchema().toString());
```
sagarkapare: nit: remove debug logs.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -61,6 +65,20 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+
+
+    if (tables.containsKey(tableName)) {
+      // TODO support same table name in different schema
+      if (!Objects.equals(tables.get(tableName), tableDetail.getSchemaName())) {
+        problems.add(new Problem("Duplicate Table Name", String.format("Duplicated table name %s detected. BigQuery " +
+          "doesn't support duplicate table name even though they are in different schema", tableName),
+          "Please only select one of the tables with same table name to replicate",
+          "Not be able to replicate multiple tables with same name " + "to BigQuery"));
+      }
```
sagarkapare, line 79: should we return from here? Since this case is not supported, no need to check additional details such as primary keys for example.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -61,6 +65,20 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
...
+      // else if schema name is same, that means it's assessing the same table.
+    } else {
+      tables.put(tableName, tableDetail.getSchemaName());
```
sagarkapare: This `else` block can be removed if we return from `if` on problem.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
```diff
@@ -61,6 +65,20 @@ public TableAssessment assess(StandardizedTableDetail tableDetail) {
     List<Problem> problems = new ArrayList<>();
     String dbName = tableDetail.getDatabase();
     String tableName = tableDetail.getTable();
+
+
+    if (tables.containsKey(tableName)) {
+      // TODO support same table name in different schema
+      if (!Objects.equals(tables.get(tableName), tableDetail.getSchemaName())) {
+        problems.add(new Problem("Duplicate Table Name", String.format("Duplicated table name %s detected. BigQuery " +
```
sagarkapare: nit: Message can be improved a little bit to include schemaNames. 
How about 
```
Table with name '%s' found in two different schemas, '%s' and '%s'. BigQuery target requires table names to be unique across the schemas. Please select table from one of the schemas only for replication.
```

sagarkapare (commented):
Can we add unit test case for this scenario?

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
seanzhougoogle: done :)

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
seanzhougoogle: done :)

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
seanzhougoogle: I think we'd better expose all the problems. Because we don't know which table the user finally will choose. For example, if user choose schema1.table1 and schema2.table1 , if we return here , we will report all the problem for schema1.table1 and only report duplicate table name problem for schema2.table2 which is not consistent. 

Reporting all the problem will have below benefits:
1) Give user more information to help them decide which table to choose finally
2) User correct the problem before they choose that table (the second table)  instead of choose the table and see new problems.

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
seanzhougoogle: this `else` corresponds to `if (tables.containsKey(tableName))`

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: done :)

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: done :)

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
nitinmotgi: I 100% agree wtih Sean. We should expose all problems. Think of the user, at least they will know the issue. If we can also suggest corrective actions that will provide an experience far better than just silently ignoring it or conveying problem that user has no clue of.

seanzhougoogle:
> Can we add unit test case for this scenario?

done :)

src/main/java/io/cdap/delta/bigquery/BigQueryAssessor.java
sagarkapare: +1 to the consistency point. Thanks for explanation Sean! Agree even this case is not supported its better to expose all the problems.

sagarkapare (approved):

PR #72, opened 2020-12-21
CDAP-17371 Propagate source database schema name to the applications setTableError method.

sagarkapare:


src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -76,7 +77,8 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
     DMLEvent event = sequencedEvent.getEvent();
     DMLOperation dmlOperation = event.getOperation();
     Key key = new Key(event.getOperation().getDatabaseName(), dmlOperation.getTableName());
-    TableObject tableObject = objects.computeIfAbsent(key, t -> new TableObject(event.getOperation().getDatabaseName(),
+    TableObject tableObject = objects.computeIfAbsent(key, t -> new TableObject(dmlOperation.getDatabaseName(),
```
seanzhougoogle, line 80: TableObject is representing something written to GCS and then loaded to BigQuery. While big query doesn't have a schema concept.

I think it would be btter to add schema to the Key instead of the Table Object.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -99,16 +101,17 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
         } catch (IOException e) {
           String errMsg = String.format("Error writing batch of %d changes for %s.%s to GCS",
                                         tableObject.numEvents, tableObject.dataset, tableObject.table);
-          context.setTableError(tableObject.dataset, tableObject.table,
+          context.setTableError(tableObject.dataset, tableObject.sourceDbSchemaName, tableObject.table,
```
seanzhougoogle, line 104: here it would be better get database name , schema name and table name from the key instead of from tableOjbect. Say if in the future we are to rename the dataset and table name (we already have such request : https://cdap.atlassian.net/jira/software/c/projects/CDAP/issues/CDAP-17337)

sagarkapare: currently database name and table name from key is same as that from the TableObject. That information is received from `DMLOperation` which is received from the source. Purpose of key in the map is to uniquely identify the table (which is databasename,tablename pair) while value carries more information such as events to write.

Since BigQuery target does not identify table by schema name separately there is no need to add it to the Key. There is open bug https://cdap.atlassian.net/browse/CDAP-17477 to address this issue.

Also can you clarify how taking these values from Key help for  https://cdap.atlassian.net/browse/CDAP-17337? For fixing this bug I imagine there will be more changes required (such as change in the DMLOperation API to carry original table name) and then decide what goes in TableObject vs Key.

seanzhougoogle: >currently database name and table name from key is same as that from the TableObject. 

That's current default implementation. And we have  https://cdap.atlassian.net/browse/CDAP-17337 to let user rename the dataset and table name of Bigquery.

> Purpose of key in the map is to uniquely identify the table (which is databasename,tablename pair)

Exactly, to identify a table we need database , schema and table name, it's irrelevant to whether target supports schema or not.

>while value carries more information such as events to write.

I think the events metrics is in terms of the names from source, right ? not from the target

>Since BigQuery target does not identify table by schema name separately there is no need to add it to the Key.

Bigquery cannot decide how to identify a table. It's the source that decide how table is identified. If the source supports schema, table should be identified by database + schema + table and otherwise databse + table. WHat's the issue that we  add schema to the Key and the schema value is null?

>Also can you clarify how taking these values from Key help for https://cdap.atlassian.net/browse/CDAP-17337?

Here you generate the metrics using BQ dataset name and the table name. Also they are same as database name and table name of the source based on current implementation, but once we let user to rename the dataset and table name, do you want to generate the metrics based on the target name instead of source name ? That's why I think using source database name , schema name and table name in the key makes it extensible and work in any case.

seanzhougoogle (approved):

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
sagarkapare: As discussed offline, fixing https://cdap.atlassian.net/browse/CDAP-17337 is not just the from where we get the database name, schema name, and table name for example `Key` vs `TableObject` but it involves more additions to the API and updating what we emit in the following accordingly - 

```
          context.setTableError(tableObject.dataset, tableObject.sourceDbSchemaName, tableObject.table,
```

PR #74, opened 2020-12-22
Make staging bucket name in the big query sink optional.

sagarkapare:
Also take location of staging GCS bucket as a configuration parameter.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
@@ -27,12 +27,14 @@ import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -85,21 +88,28 @@ public EventConsumer createConsumer(DeltaTargetContext context) throws IOExcepti
       .setProjectId(project)
       .build()
       .getService();
-    // TODO: make bucket optional
-    Bucket bucket = storage.get(conf.stagingBucket);
+
+    String stagingBucketName = conf.stagingBucket == null ? null : conf.stagingBucket.trim();
```
seanzhougoogle, line 92: besides trim, shall me make it to lowerCase too ? GCS bucket name only allow lower case

sagarkapare: yeah making lower case on line 96 below.

seanzhougoogle (approved):

PR #78, opened 2020-12-30
Fix the duplicate events issue

seanzhougoogle:
issue : duplicate insert events merged to target table

root cause : 
we didn't clear staging table once there are errors. Thus if a `flush` fails and some of the events were already loaded to staging table and the offset is not committed. Once delta app resume from last Offset,  duplicate insert events are loaded to the same staging table thus got merged to target table.

solution :
1)  clear the staging table each time there is a failure.
2) clear the staging table before loading GCS blob to staging table

Note:
uncleared GCS staging area won't cause this issue, because each time we load the GCS file that identified by a batch Id which will be different for each `flush`


seanzhougoogle:
close the PR since uncleared staging table should not cause issue because we filter events by batch Id in diff query and batch Id will be different for each 'flush'

PR #80, opened 2020-12-30
Use source time stamp to decide the ordering when source events are unordered

seanzhougoogle:


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -481,6 +485,7 @@ private Schema addSupplementaryColumnsToTargetSchema(Schema original) {
     fields.add(Schema.Field.of(Constants.SEQUENCE_NUM, Schema.of(Schema.Type.LONG)));
     fields.add(Schema.Field.of(Constants.IS_DELETED, Schema.nullableOf(Schema.of(Schema.Type.BOOLEAN))));
     fields.add(Schema.Field.of(Constants.ROW_ID, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+    fields.add(Schema.Field.of(Constants.SOURCE_TIMESTAMP, Schema.nullableOf(Schema.of(Schema.Type.LONG))));
```
sagarkapare, line 488: Line 484 should be now
```
List<Schema.Field> fields = new ArrayList<>(original.getFields().size() + 4);
```
to take in account for the newly added SOURCE_TIMESTAMP field.

sagarkapare (approved):
One comment else LGTM.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: good catch :)

sagarkapare: Also can you update documentation in this class accordingly to reflect these changes?

seanzhougoogle: done

PR #88, opened 2021-01-10
insert a deleted row for unordered delete event that doesn't match any exisitng row and add more debugging logs

seanzhougoogle:
What:
Insert a row with "_is_deleted = true" for an unordered delete event that doesn't match any existing row 

Why:
For unordered events, when a delete event is not matching any row in the target table, that means both insert and delete to the same row are in the same batch and thus diff query will only get the delete event in the result, we should insert a row with _is_delete as true, because it's possible there is another update event to the same row that will arrive late in next batch and if we don't have this row here this late arriving update event will insert a new row.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -317,7 +317,7 @@ public class BigQueryEventConsumer implements EventConsumer {
 
     latestOffset = event.getOffset();
     latestSequenceNum = sequenceNumber;
-    LOG.debug("DDL offset: {} seq num: {}", latestOffset.get(), latestSequenceNum);
+    LOG.trace("DDL offset: {} seq num: {}", latestOffset.get(), latestSequenceNum);
     context.incrementCount(event.getOperation());
     if (event.isSnapshot()) {
       context.setTableSnapshotting(normalizedDatabaseName, normalizedTableName);
@@ -512,7 +512,7 @@ public class BigQueryEventConsumer implements EventConsumer {
     try {
       Failsafe.with(commitRetryPolicy).run(() -> {
         if (latestOffset != null) {
-          LOG.debug("Committing offset : {} and seq num: {}" , latestOffset.get(), latestSequenceNum);
+          LOG.trace("Committing offset : {} and seq num: {}" , latestOffset.get(), latestSequenceNum);
           context.commitOffset(latestOffset, latestSequenceNum);
         }
       });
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -557,6 +559,7 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
 
     latestOffset = event.getOffset();
     latestSequenceNum = sequenceNumber;
+    LOG.debug("DML id:{} offset: {} seq num : {}", event.getRow().get("id"), latestOffset.get(), latestSequenceNum);
```
sagarkapare: nit: remove these debug logs.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -511,6 +512,7 @@ private void commitOffset() throws DeltaFailureException {
     try {
       Failsafe.with(commitRetryPolicy).run(() -> {
         if (latestOffset != null) {
+          LOG.debug("Committing offset : {} and seq num: {}" , latestOffset.get(), latestSequenceNum);
```
sagarkapare: nit: remove these debug logs.

sagarkapare (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -317,6 +317,7 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
 
     latestOffset = event.getOffset();
     latestSequenceNum = sequenceNumber;
+    LOG.debug("DDL offset: {} seq num: {}", latestOffset.get(), latestSequenceNum);
```
sagarkapare: nit: remove this.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: as discussed offline, change it to trace level.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: as discussed offline, change it to trace level.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: as discussed offline, change it to trace level.

PR #91, opened 2021-01-14
handle events with same timestamp against the same row more properly

seanzhougoogle:
When for the same rowid source timestamp are same late coming event (with larger sequence number ) should be considered as greater order

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
+++ b/src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
@@ -20,6 +20,7 @@ import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Storage;
+import com.google.gson.Gson;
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -226,7 +231,11 @@ private void writeEvent(Sequenced<DMLEvent> sequencedEvent) throws IOException {
         avroWriter = new DataFileWriter<>(datumWriter).create(avroSchema, outputStream);
       }
 
-      avroWriter.append(createStagingRecord(sequencedEvent));
+      StructuredRecord stagingRecord = createStagingRecord(sequencedEvent);
+      if (LOG.isTraceEnabled()) {
+        LOG.error("Writing record {} to GCS.", GSON.toJson(stagingRecord));
```
sagarkapare: nit: LOG.trace()

seanzhougoogle: done :)

sagarkapare (approved):

PR #93, opened 2021-01-25
update _is_deleted to null when merge insert/update events

seanzhougoogle:


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -1101,7 +1101,7 @@ static String createMergeQuery(TableId targetTableId, List<String> primaryKeys,
         .filter(predicate)
         .map(Schema.Field::getName)
         .map(name -> String.format("%s = D.%s", name, name))
-        .collect(Collectors.joining(", ")) + "\n" +
+        .collect(Collectors.joining(", ")) +  ", " + Constants.IS_DELETED + " = null\n" +
```
sagarkapare, line 1107: Do we need to explicitly set to `null`? this field should be `NULLABLE` right?

seanzhougoogle: yes, this filed is nullable. This is for the case when this row was previously deleted and the "_is_deleted" column was set to "true" and now a new insert is to insert the same row , we need to reset "_is_deleted" back to null.

sagarkapare: I see. Thanks for the explanation! Can we add comment describing why we need to set `IS_DELETED` explicitly to `null`?

sagarkapare (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: done :)

PR #95, opened 2021-02-09
CDAP-17674 support runtime arguments for retain staging table

seanzhougoogle:
What:
let user to specify a runtime argument that will retain the staging table
Why:
when there is an issue we can run diff query on staging table for debugging

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -164,6 +164,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   private static final String VALID_NAME_REGEX = "[\\w]+";
   private static final String INVALID_NAME_REGEX = "[^\\w]+";
   private static final Gson GSON = new Gson();
+  private static final String RETAIN_STAGING_TABLE = "table.staging.retain";
 
   private final DeltaTargetContext context;
   private final BigQuery bigQuery;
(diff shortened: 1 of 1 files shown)


albertshau (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -164,6 +164,7 @@
   private static final String VALID_NAME_REGEX = "[\\w]+";
   private static final String INVALID_NAME_REGEX = "[^\\w]+";
   private static final Gson GSON = new Gson();
+  private static final String RETAIN_STAGING_TABLE = "table.staging.retain";
```
sagarkapare: nit: `"retain.staging.table"` similar to how variable is named.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -236,6 +238,8 @@
     this.sourceEventOrdering = context.getSourceProperties() == null ? SourceProperties.Ordering.ORDERED :
       context.getSourceProperties().getOrdering();
     this.datasetName = datasetName;
+    this.retainStagingTable = Boolean.parseBoolean(context.getRuntimeArguments().getOrDefault(RETAIN_STAGING_TABLE,
```
sagarkapare: minor nit: no need to assign default. `null` value is treated as `false`. so could be simplified as 
```
this.retainStagingTable = Boolean.parseBoolean(context.getRuntimeArguments().get(RETAIN_STAGING_TABLE));
```

sagarkapare (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: done :)

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
seanzhougoogle: done:)

PR #97, opened 2021-02-18
CDAP-17601 Load data directly into the target table during snapshot process.

sagarkapare:
JIRA - https://cdap.atlassian.net/browse/CDAP-17601

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -155,6 +155,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryEventConsumer.class);
   private static final Gson GSON = new Gson();
   private static final String RETAIN_STAGING_TABLE = "retain.staging.table";
+  private static final String DIRECT_LOADING_IN_PROGRESS_PREFIX = "bigquery-direct-load-in-progress-";
 
   private final DeltaTargetContext context;
   private final BigQuery bigQuery;
@@ -348,6 +349,14 @@ public class BigQueryEventConsumer implements EventConsumer {
       case CREATE_TABLE:
         TableId tableId = TableId.of(project, normalizedDatabaseName, normalizedTableName);
         Table table = bigQuery.getTable(tableId);
+        // SNAPSHOT data is directly loaded in the target table. Check if any such direct load was in progress
+        // for the current table when target received CREATE_TABLE ddl. This indicates that the snapshot was abandoned
+        // because of some failure scenario. Delete the existing table if any.
+        byte[] state = context.getState(String.format(DIRECT_LOADING_IN_PROGRESS_PREFIX + "%s-%s",
+                                                      normalizedDatabaseName, normalizedTableName));
+        if (table != null && state != null && Bytes.toBoolean(state)) {
+          bigQuery.delete(tableId);
+        }
         List<String> primaryKeys = event.getPrimaryKey();
         updatePrimaryKeys(tableId, primaryKeys);
         // TODO: check schema of table if it exists already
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -69,23 +70,23 @@ public MultiGCSWriter(Storage storage, String bucket, String baseObjectName,
     this.storage = storage;
     this.bucket = bucket;
     this.baseObjectName = baseObjectName;
-    this.objects = new HashMap<>();
+    this.objects = new LinkedHashMap<>();
```
sagarkapare: This change is made to make sure that snapshot records landed first in the target. `LinkedHashMap` will provide insert ordering guarantee and since snapshot records come in first, key with the snapshot enabled will be first.

albertshau: if ordering is important it's better to use a List, otherwise it will be easy to break this accidentally in the future.

albertshau: though the key already has snapshot info, it would be more clear to explicitly look for that information to load snapshots first rather than relying on order.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -619,11 +623,30 @@ synchronized void flush() throws InterruptedException, IOException, DeltaFailure
     commitOffset();
   }
 
+  private void directLoadToTarget(TableBlob blob) throws InterruptedException, DeltaFailureException {
```
albertshau: Actually, this can result in duplicate data if there is a failure during the snapshot processing and we reprocess events. To protect against this, we would have to write some state to remember if this is the first time, or wipe out the existing data if it already exists.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
sagarkapare: Added enum to distinguish between `SNAPSHOT` and `STREAMING` blobs. PTAL.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sagarkapare: Added state to indicate whether direct load was in progress per table. If such load was in progress and we receive CREATE_TABLE ddl again, it indicates that the snapshot was abandoned. Clearing table before proceeding in such cases.

albertshau: Are there use cases for writing to pre-existing tables? If there are, clearing the table may not be desired, though I'm not sure when you would ever want to snapshot to a pre-existing table.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -568,22 +577,42 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
...
+
+  private void processBlobsInParallel(Collection<TableBlob> tableBlobs)
+    throws InterruptedException, DeltaFailureException {
     List<Future<?>> mergeFutures = new ArrayList<>(tableBlobs.size());
     for (TableBlob blob : tableBlobs) {
       // submit a callable instead of a runnable so that it can throw checked exceptions
       mergeFutures.add(executorService.submit((Callable<Void>) () -> {
-        mergeTableChanges(blob);
+        if (blob.isSnapshotOnly()) {
+          context.putState(String.format(DIRECT_LOADING_IN_PROGRESS_PREFIX + "%s-%s", blob.getDataset(),
```
albertshau, line 606: does this get cleared or set to false anywhere?

sagarkapare: yes it does get set to false in the else block below on line number 611

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sagarkapare: Yes. There is use case of writing to the pre-existing tables. However in this case user would want to select the `skip  snapshot` option available from source plugins. When this option is selected `directLoadToTarget` wont get called since it's applicable to snapshot records only. 

However I am not aware of any use case where user wants to write to pre-existing table with snapshot data. @bdmogal can comment on this. However in this case too when `DROP TABLE` is generated by source, target can delete the existing table if `requireManualDrop` is set to false which is default.

albertshau (approved):

sagarkapare:
https://builds.cask.co/browse/DEL-BPBT94 Build is green. Merging.

PR #98, opened 2021-02-24
CDAP-17482 Compute and set the sequence numbers based on the content of existing bq tables.

sagarkapare:
JIRA: https://cdap.atlassian.net/browse/CDAP-17482

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.delta.bigquery;
+
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.EncryptionConfiguration;
+import com.google.cloud.bigquery.FieldValue;
+import com.google.cloud.bigquery.FieldValueList;
+import com.google.cloud.bigquery.Job;
+import com.google.cloud.bigquery.JobId;
+import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.TableId;
+import com.google.cloud.bigquery.TableResult;
+import io.cdap.delta.api.SourceTable;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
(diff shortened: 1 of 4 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,130 @@
...
+                                               EncryptionConfiguration encryptionConfig) throws Exception {
+    if (bigQuery.getTable(tableId) == null) {
+      return 0L;
+    }
+
+    String query = String.format("SELECT MAX(_sequence_num) FROM %s.%s", tableId.getDataset(), tableId.getTable());
+    return executeMaxQuery(bigQuery, query, encryptionConfig);
+  }
+
+  private static long executeMaxQuery(BigQuery bigQuery, String query, EncryptionConfiguration encryptionConfig)
```
seanzhougoogle: nit: executeAggregateQuery

Here the function itself is not executing a MaxQuery, instead it's executing a query passed in which is expected to only generate an aggregated result (one row and one column)

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,130 @@
...
+    JobId jobId = JobId.of(UUID.randomUUID().toString());
+    Job queryJob = bigQuery.create(JobInfo.newBuilder(jobConfig).setJobId(jobId).build());
+    queryJob.waitFor();
+    TableResult result = queryJob.getQueryResults();
+    Iterator<FieldValueList> resultIter = result.iterateAll().iterator();
+    if (!resultIter.hasNext()) {
+      return 0L;
+    }
+    // query is SELECT MAX(_sequence_num) FROM ...
+    // so there is at most one row and one column in the output.
```
seanzhougoogle: nit:
// the query passed in is expected to generate an aggregated result which contains one row and one column.

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -0,0 +1,130 @@
...
+    builder.append("SELECT MAX(max_sequence_num) FROM (");
+    Set<SourceTable> allTables = context.getAllTables();
+    List<String> maxSequenceNumQueryPerTable = new ArrayList<>();
+    for (SourceTable table : allTables) {
+      TableId tableId = TableId.of(project,
+                                   datasetName != null ? normalize(datasetName) : normalize(table.getDatabase()),
+                                   normalize(table.getTable()));
+      if (bigQuery.getTable(tableId) != null) {
+        maxSequenceNumQueryPerTable.add(String.format("SELECT MAX(_sequence_num) as max_sequence_num FROM %s.%s.%s",
+                                                      tableId.getProject(), tableId.getDataset(), tableId.getTable()));
```
seanzhougoogle: do we need the project ? getMaximumSequenceNumberForTable does't use project, shall we make them consistent?

seanzhougoogle (approved):
Leave some minor comments otherwise LGTM :)

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
sagarkapare: +1 fixed.

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
sagarkapare: fixed.

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
sagarkapare: Its not required. Removed.

sagarkapare:
https://builds.cask.co/browse/DEL-BPBT95-4 Squashing and merging.

PR #111, opened 2021-04-15
PLUGIN-645 Add support for DateTime in replication BQ target.

sagarkapare:
https://cdap.atlassian.net/browse/PLUGIN-645
This PR has following changes:
1. `EventWriter` interface and it's implementations `AvroEventWriter` and `JsonEventWriter`. `JsonEventWriter` is used when table is determined to have `DateTime` field in it, else `AvroEventWriter` is used to write StructuredRecord to GCS bucket.
2. Copy of class https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQueryRecordToJson.java and renamed to `StructuredRecordToJson`.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/AvroEventWriter.java
+++ b/src/main/java/io/cdap/delta/bigquery/AvroEventWriter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.delta.bigquery;
+
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import org.apache.avro.file.DataFileWriter;
+import org.apache.avro.io.DatumWriter;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * EventWriter that writes records in Avro format.
+ */
+public class AvroEventWriter implements EventWriter {
+  private DataFileWriter<StructuredRecord> avroWriter;
+
+  AvroEventWriter(org.apache.avro.Schema avroSchema, OutputStream outputStream) {
+    DatumWriter<StructuredRecord> datumWriter = new RecordDatumWriter();
+    try {
+      avroWriter = new DataFileWriter<>(datumWriter).create(avroSchema, outputStream);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to create avro event writer.", e);
+    }
+  }
+
+  @Override
+  public void write(StructuredRecord record) throws IOException {
+    avroWriter.append(record);
+  }
+
+  @Override
+  public void close() throws IOException {
+    avroWriter.close();
+  }
+}
--- a/src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
+++ b/src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
@@ -29,8 +29,6 @@ import io.cdap.delta.api.DeltaTargetContext;
 import io.cdap.delta.api.ReplicationError;
 import io.cdap.delta.api.Sequenced;
 import io.cdap.delta.api.SourceProperties;
-import org.apache.avro.file.DataFileWriter;
-import org.apache.avro.io.DatumWriter;
 import org.slf4j.Logger;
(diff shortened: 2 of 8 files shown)


src/main/java/io/cdap/delta/bigquery/AvroEventWriter.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
greeshmaswaminathan: 2021 for all the files

src/main/java/io/cdap/delta/bigquery/StructuredRecordToJson.java
```diff
@@ -0,0 +1,298 @@
...
+      case DOUBLE:
+        writer.value((Number) object);
+        break;
+      case BOOLEAN:
+        writer.value((Boolean) object);
+        break;
+      case STRING:
+        writer.value(object.toString());
+        break;
+      case BYTES:
```
greeshmaswaminathan, line 172: This section was newly added in 6.4. Please do some additional testing with bytes to ensure it is working as expected for replication (Not anticipating any issue , but just to be sure).

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -103,6 +102,20 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
     }
   }
 
+  private boolean isJsonFormat(StructuredRecord record) {
+    List<Schema.Field> fields = record.getSchema().getFields();
+    if (fields == null) {
+      return false;
+    }
+    for (Schema.Field field : fields) {
+      Schema.LogicalType logicalType = field.getSchema().getLogicalType();
```
greeshmaswaminathan: getNonNullableSchema here if it is possible to specify field as nullable

src/main/java/io/cdap/delta/bigquery/AvroEventWriter.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class AvroEventWriter implements EventWriter {
+  private DataFileWriter<StructuredRecord> avroWriter;
+
+  AvroEventWriter(org.apache.avro.Schema avroSchema, OutputStream outputStream) {
+    DatumWriter<StructuredRecord> datumWriter = new RecordDatumWriter();
+    try {
+      avroWriter = new DataFileWriter<>(datumWriter).create(avroSchema, outputStream);
+    } catch (IOException e) {
+      e.printStackTrace();
```
seanzhougoogle: do we need to handle the exception here by logging the issue and throw a runtime exception ? otherwise avroWriter will be null

src/main/java/io/cdap/delta/bigquery/AvroEventWriter.java
sagarkapare: fixed.

src/main/java/io/cdap/delta/bigquery/StructuredRecordToJson.java
sagarkapare: ack.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
sagarkapare: fixed.

src/main/java/io/cdap/delta/bigquery/AvroEventWriter.java
sagarkapare: ah.. forgot to change it. yeah we should throw exception from here. fixed the code.

seanzhougoogle (approved):

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
```diff
@@ -103,6 +102,20 @@ public synchronized void write(Sequenced<DMLEvent> sequencedEvent) {
     }
   }
 
+  private boolean isJsonFormat(StructuredRecord record) {
+    List<Schema.Field> fields = record.getSchema().getFields();
+    if (fields == null) {
+      return false;
+    }
+    for (Schema.Field field : fields) {
+      Schema.LogicalType logicalType = field.getSchema().getNonNullable().getLogicalType();
```
greeshmaswaminathan: This may be null for fields that are not nullable. Typically I see checks like `isNullable ? field.getSchema().getNonNullable().getType() : field.getSchema().getType()`. https://github.com/cdapio/cdap/blob/aa90c10ed4a7e0638f5d57db52b8d8e41830738c/cdap-api-common/src/main/java/io/cdap/cdap/api/data/format/StructuredRecord.java#L574

greeshmaswaminathan (approved):
Just one comment on the nullable schema check. LGTM otherwise.

sagarkapare:
Build is green. https://builds.cask.co/browse/DEL-BPBT108-3 Squashing and merging.

src/main/java/io/cdap/delta/bigquery/MultiGCSWriter.java
sagarkapare: fixed.

PR #116, opened 2021-05-03
CDAP-17720 normalize column name

seanzhougoogle:


src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -121,8 +127,10 @@ private static long executeAggregateQuery(BigQuery bigQuery, String query, Encry
 
     return val.getLongValue();
   }
-
   public static String normalize(String name) {
+    return normalize(name, MAX_LENGTH);
+  }
+  public static String normalize(String name, int maxLength) {
```
sagarkapare: Would be good to add javadoc to this method explaining how we normalize the names.

seanzhougoogle: thanks for the suggestion:) That also helped me find a bug in my implementation when comparing the java doc with my impl :)

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -134,9 +160,47 @@ public static String normalize(String name) {
...
-      name = name.substring(0, MAX_LENGTH);
+    if (name.length() > maxLength) {
+      name = name.substring(0, maxLength);
+    }
+
+    // replace the first character with underscore if it's a number and the name cannot start with number
+    if (!canStartWithNumber) {
+      char first = name.charAt(0);
+      if (first >= '0' && first <= '9') {
+        name = "_" + name.substring(1);
```
sagarkapare: why cant we just prepend with `_` in this case rather than replacing with first digit?

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -134,9 +160,47 @@ public static String normalize(String name) {
     }
 
     // truncate the name if it exceeds the max length
-    if (name.length() > MAX_LENGTH) {
-      name = name.substring(0, MAX_LENGTH);
+    if (name.length() > maxLength) {
+      name = name.substring(0, maxLength);
+    }
+
+    // replace the first character with underscore if it's a number and the name cannot start with number
```
sagarkapare: nit: replace the first character with underscore if it's a number as the name cannot start with number

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -122,10 +127,31 @@ private static long executeAggregateQuery(BigQuery bigQuery, String query, Encry
...
+   * And it must be 128 characters or fewer.
+   * @param name the field name to be normalized
+   * @return the normalized name
+   */
+  public static String normalizeFieldName(String name) {
+    return normalize(name, FIELD_NAME_MAX_LENGTH, false);
+  }
+
+  private static String normalize(String name, int maxLength, boolean canStartWithNumber) {
+    if (name == null || name.isEmpty()) {
```
sagarkapare, line 153: when will `name` be null?

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    assertEquals(name.substring(0, 1024), BigQueryUtils.normalizeDatasetOrTableName(name));
+    // contains invalid character
+    assertEquals("ab_c", BigQueryUtils.normalizeDatasetOrTableName("ab?/c"));
+
+  }
+
+  @Test
+  public void testNormalizeFieldName() {
+    // only contains number and letter
+    assertEquals("a2fs", BigQueryUtils.normalizeFieldName("a2fs"));
```
sagarkapare, line 49: Let's add test case with space in the field name too? for example "database version".

sagarkapare (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
seanzhougoogle: done :)

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
seanzhougoogle: revised it a little bit different, because whether the name can start with number depends the flag "canStartWithNumber"

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
seanzhougoogle: currently it should not happen. I retain the original logic, I think it's just for safety.

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
seanzhougoogle: done :)

PR #118, opened 2021-05-19
CDAP-18002 persist direct loading in progress state in boolean

seanzhougoogle:
the state was read as boolean and written as string, which caused below exception 
java.lang.IllegalArgumentException: Array has wrong size: 4
	at io.cdap.cdap.api.common.Bytes.toBoolean(Bytes.java:415) ~[na:na]
	at io.cdap.delta.bigquery.BigQueryEventConsumer.handleDDL(BigQueryEventConsumer.java:367) ~[1621448876368-

make them consistent


avirmani (approved):
Did we not have a test covering boolean on the source?

PR #123, opened 2021-06-28
CDAP-17895 : making batches of 1000 tables for each call to big query and accumulating later

sahusanket:


src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -64,6 +64,7 @@
   public static final int CONFLICT = 409;
   private static final String GCS_SCHEME = "gs://";
   private static final String GCP_CMEK_KEY_NAME = "gcp.cmek.key.name";
+  private static final int MAX_BIG_QUERY_BATCH_SZE = 1000;
```
sagarkapare: nit: maybe call it as MAX_TABLES_PER_QUERY?

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -60,9 +60,8 @@ private BigQueryUtils() {
    * missing '_sequence_num' column in any of the tables, exception will be thrown.
    */
   static long getMaximumExistingSequenceNumber(Set<SourceTable> allTables, String project, @Nullable String datasetName,
-                                               BigQuery bigQuery, EncryptionConfiguration encryptionConfiguration) {
-    StringBuilder builder = new StringBuilder();
-    builder.append("SELECT MAX(max_sequence_num) FROM (");
+                                               BigQuery bigQuery, EncryptionConfiguration encryptionConfiguration,
+                                               int splitBatchSize) {
```
sagarkapare: nit: int maxTablesPerQuery

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -60,9 +60,8 @@ private BigQueryUtils() {
    * missing '_sequence_num' column in any of the tables, exception will be thrown.
    */
   static long getMaximumExistingSequenceNumber(Set<SourceTable> allTables, String project, @Nullable String datasetName,
```
sagarkapare: Logic seems little bit complicated. We already have method that gives max sequence number from Set of tables (`allTables`), so we just have to make sure that this set will have max 1000 tables. something similar to below

```java
    Set<SourceTable> allRemainingTables = context.getAllTables();
    long maxExisting = 0;
    while (allRemainingTables.size() > 0) {
     // select 1000 tables from all tables
      HashSet<SourceTable> currentBatch = Sets.newHashSet(Iterables.limit(allRemainingTables, 1000));
     // find max among those 1000 tables
      long currentMax = BigQueryUtils.getMaximumExistingSequenceNumber(currentBatch, project, 
                                                                       conf.getDatasetName(), bigQuery, 
                                                                       encryptionConfig);
      maxExisting = Math.max(maxExisting, currentMax);
     // remove current batch of tables from all tables.
      allRemainingTables = Sets.difference(allRemainingTables, currentBatch);
    }
    
    return maxExisting;
```

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
```diff
@@ -17,50 +17,315 @@
...
+                                                                         null, bigQueryMock, null, 1000);
+      assertEquals(0L, tableResult0);
+      PowerMockito.verifyPrivate(BigQueryUtils.class, times(0))
+        .invoke("executeAggregateQuery",
+                ArgumentMatchers.eq(bigQueryMock), ArgumentMatchers.any(), ArgumentMatchers.any());
+
+    }
+
+    @Test
+    public void testGetMaximumExistingSequenceNumberSingleInvocations() throws Exception {
```
sagarkapare, line 147: We dont need this test now right? or anything related to PowerMockito, since we will directly executing test against bq?

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
sahusanket: done

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
sahusanket: Added an extra method to handle this logic

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
sahusanket: I was thinking these are required because : 
i. These tests will run at least, incase GCP service role is not provided
ii. These tests actually checks how many invocations/calls will go to Big query depending on size of table list

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -92,9 +93,9 @@ public void initialize(DeltaTargetContext context) throws Exception {
       .build()
       .getService();
 
-    long maximumExistingSequenceNumber = BigQueryUtils.getMaximumExistingSequenceNumber(context.getAllTables(), project,
-                                                                                        conf.getDatasetName(), bigQuery,
-                                                                                        encryptionConfig);
+    long maximumExistingSequenceNumber =
+      BigQueryUtils.getMaximumExistingSequenceNumberBatchSpliter(context.getAllTables(), project, conf.getDatasetName(),
```
sagarkapare: nit: probably we can call this method as `getMaximumExistingSequenceNumber`

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -55,6 +58,28 @@
...
+                                                           EncryptionConfiguration encryptionConfiguration,
+                                                           int maxTablesPerQuery) {
+
+    Set<SourceTable> allRemainingTables = allTables;
+    long maxExisting = 0;
+    while (allRemainingTables.size() > 0) {
+      // select 1000 tables from all tables
+      HashSet<SourceTable> currentBatch = Sets.newHashSet(Iterables.limit(allRemainingTables, maxTablesPerQuery));
+      // find max among those 1000 tables
+      long currentMax = BigQueryUtils.getMaximumExistingSequenceNumber(currentBatch, project,
```
sagarkapare: nit: then we call this method `getMaximumExistingSequenceNumberPerBatch`

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -55,6 +58,28 @@
...
 
+  static long getMaximumExistingSequenceNumberBatchSpliter(Set<SourceTable> allTables, String project,
+                                                           @Nullable String datasetName, BigQuery bigQuery,
+                                                           EncryptionConfiguration encryptionConfiguration,
+                                                           int maxTablesPerQuery) {
+
+    Set<SourceTable> allRemainingTables = allTables;
+    long maxExisting = 0;
+    while (allRemainingTables.size() > 0) {
+      // select 1000 tables from all tables
```
sagarkapare: nit: select `maxTablesPerQuery` tables from all tables.

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -55,6 +58,28 @@
...
+      HashSet<SourceTable> currentBatch = Sets.newHashSet(Iterables.limit(allRemainingTables, maxTablesPerQuery));
+      // find max among those 1000 tables
```
sagarkapare: nit: find max among those `maxTablesPerQuery` tables.

sagarkapare (approved):
minor nits. else LGTM 👍 
Please squash commits into one before merging.

PR #127, opened 2021-07-07
Set the type in BigQuery as BigNumeric when scale is more than 9 for NUMERIC.

sagarkapare:


src/main/java/io/cdap/delta/bigquery/Schemas.java
```diff
@@ -113,7 +122,7 @@ private static Field convertToBigQueryField(Schema.Field field) {
     Schema.LogicalType logicalType = fieldSchema.getLogicalType();
     Field.Mode fieldMode = isNullable ? Field.Mode.NULLABLE : Field.Mode.REQUIRED;
     if (logicalType != null) {
-      StandardSQLTypeName bqType = convertLogicalType(logicalType);
+      StandardSQLTypeName bqType = convertLogicalType(fieldSchema);
```
seanzhougoogle, line 125: why do we need to change the method signature from accepting a logicalType to accepting a Schema ? if the method is meant for converting logical type ?

seanzhougoogle (approved):
LGTM with one minor comment.

src/main/java/io/cdap/delta/bigquery/Schemas.java
sagarkapare: We need additional details while converting logical type such as `precision` and `scale`. These details are not present in the `Schema.LogicalType` but present in the fieldschema.

sagarkapare:
https://builds.cask.co/browse/DEL-BPBT124-2 build is green. Merging.

PR #131, opened 2021-07-16
CDAP-18194 Remove un-necessary restriction of not allowing BigNumeric…

sagarkapare:
… fields when using JSON format.


seanzhougoogle (approved):

sagarkapare:
https://builds.cask.co/browse/DEL-BPBT128

PR #138, opened 2021-09-07
Set flag to use Avro Logical Types in replication pipelines

fernst:


seanzhougoogle (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -739,6 +739,7 @@ private void loadTable(TableId tableId, TableBlob blob, boolean directLoadToTarg
       = Schemas.convert(directLoadToTarget ? blob.getTargetSchema() : blob.getStagingSchema());
     LoadJobConfiguration.Builder jobConfigBuilder = LoadJobConfiguration.newBuilder(tableId, uri)
       .setFormatOptions(FormatOptions.avro())
+      .setUseAvroLogicalTypes(true)
```
sagarkapare, line 742: On line 748 when blob format is JSON we set the format options to json.
`jobConfigBuilder.setFormatOptions(FormatOptions.json());`

Should this be only done when blob is not JSON format? 
i.e. 

```java
LoadJobConfiguration.Builder jobConfigBuilder = LoadJobConfiguration.newBuilder(tableId, uri);
if (blob.isJsonFormat()) {
   // use JSON format
      jobConfigBuilder.setFormatOptions(FormatOptions.json());
} else {
// avro
jobConfigBuilder.setFormatOptions(FormatOptions.json());
.setUseAvroLogicalTypes(true)
}
```

seanzhougoogle: agree, we should also make line 741 `setFormatOptions(FormatOptions.avro())` conditional

PR #154, opened 2022-02-09
CDAP-18456 Allow option to soft delete for BQ target.

sagarkapare:
BUG: https://cdap.atlassian.net/browse/CDAP-18456

### DIFF
--- a/docs/bigquery-cdcTarget.md
+++ b/docs/bigquery-cdcTarget.md
@@ -76,3 +76,7 @@ any characters exceeds length limit will be truncated.
 **Encryption Key Name**: GCP Customer-managed encryption key (CMEK) used to encrypt the resources created by this target. 
 Encryption key name should be of the form "projects/<gcp-project-id>/locations/<key-location>/keyRings/<key-ring-name>/cryptoKeys/<key-name>".
 
+**Enable Soft Deletes**: Determine whether to enable soft deletes. If set to true, when the delete event
+is received by the target, '_is_deleted' column for the record will be set to true, otherwise record will be deleted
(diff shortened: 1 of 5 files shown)


sahusanket (approved):

docs/bigquery-cdcTarget.md
```diff
@@ -76,3 +76,7 @@ any characters exceeds length limit will be truncated.
 **Encryption Key Name**: GCP Customer-managed encryption key (CMEK) used to encrypt the resources created by this target. 
 Encryption key name should be of the form "projects/<gcp-project-id>/locations/<key-location>/keyRings/<key-ring-name>/cryptoKeys/<key-name>".
 
+**Enable Soft Deletes**: Determine whether to enable soft deletes. If set to true, when the delete event
+is received by the target, '_is_deleted' column for the record will be set to true, otherwise record will be deleted
+ from the BigQuery table. This configuration is no-op for the sources which generates events out of order, in which
```
seanzhougoogle, line 81: maybe we should explicitly callout which source is generating events out of order (oracle via datastream), because customer is not aware of that

seanzhougoogle: or maybe we should add doc to each source indicating whether that source is generating events in order or out of order

seanzhougoogle (approved):

docs/bigquery-cdcTarget.md
sagarkapare: Good point! Would be good to add it in the source documentation though since BQ target does have knowledge of all sources.

sagarkapare:
Build link https://builds.cask.co/browse/DEL-BPBT153-2

PR #171, opened 2022-06-22
CDAP-19180 :  change doc md as per suggestion

sahusanket:
Changing as per suggestion in https://cdap.atlassian.net/browse/CDAP-19180


seanzhougoogle (commented):
Talked to Bhooshan, looks like we don't need this change

PR #172, opened 2022-07-25
Updated table name validation to work with all valid characters

adinilfeld:
Previously, BigQuery table names could only contain letters, numbers, and underscores. The valid character set has since been expanded ([reference](https://cloud.google.com/bigquery/docs/tables#table_naming)), so normalization currently unnecessarily changes some otherwise valid table names.

This PR fixes this issue by updating the regex used to normalize table names. I also added relevant unit tests.

Relevant Jira issue: [PLUGIN-1325](https://cdap.atlassian.net/browse/PLUGIN-1325).

[PR #1090](https://github.com/data-integrations/google-cloud/pull/1090) in `google-cloud` is related.


fernst (approved):
LGTM

sechegaray (approved):

PR #173, opened 2022-08-18
Bug/cdap 19442 - normalize column names in big query schema 

jingtaowu3:
This PR is opened to address JIRA [CDAP-19442](https://cdap.atlassian.net/browse/CDAP-19442)

Currently normalization is not applied to column names and clustering fields, so we will see an exception when creating table with invalid format (eg, column name that as dash in it) similarly during replication job tries to create such tables. The way to fix this is to apply the existing BigQueryUtils.normalizeFieldName when converting schema to BigQuery schema, so that all fields are in line with BigQuery requirements.

For validation, a new unit test has been added to make sure invalid column fields/clustering fields are handled properly.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/Schemas.java
+++ b/src/main/java/io/cdap/delta/bigquery/Schemas.java
@@ -115,7 +115,7 @@ public class Schemas {
   }
 
   private static Field convertToBigQueryField(Schema.Field field) {
-    String name = field.getName();
+    String normalizedName = BigQueryUtils.normalizeFieldName(field.getName());
     boolean isNullable = field.getSchema().isNullable();
     Schema fieldSchema = field.getSchema();
     fieldSchema = isNullable ? fieldSchema.getNonNullable() : fieldSchema;
@@ -127,9 +127,9 @@ public class Schemas {
       if (bqType == null) {
         throw new IllegalArgumentException(
           String.format("Field '%s' is of type '%s', which is not supported in BigQuery.",
-                        name, logicalType.getToken()));
+                        normalizedName, logicalType.getToken()));
       }
-      return Field.newBuilder(name, bqType).setMode(fieldMode).build();
+      return Field.newBuilder(normalizedName, bqType).setMode(fieldMode).build();
     }
 
     Field output;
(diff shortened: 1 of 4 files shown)


src/test/java/io/cdap/delta/bigquery/BigQueryEventConsumerTest.java
```diff
@@ -482,6 +482,46 @@ public void testSoftDeletes() throws Exception {
...
+          .build();
+      eventConsumer.applyDDL(new Sequenced<>(createTable, 0));
+
+      Table table = bigQuery.getTable(TableId.of(dataset, normalizedTableName));
+      Assert.assertNotNull(table);
+      FieldList bqFields = table.getDefinition().getSchema().getFields();
+      Field normalizedField = bqFields.get(normalizedFieldName);
+      Assert.assertNotNull(normalizedField);
+      Assert.assertEquals(LegacySQLTypeName.INTEGER, normalizedField.getType());
+    }finally {
```
yaojiefeng: nit - add a space after `}`, checkstyle will fail. Please run `mvn clean package -DskipTests` to ensure the checkstyle passes

src/test/java/io/cdap/delta/bigquery/BigQueryEventConsumerTest.java
```diff
@@ -482,6 +482,46 @@ public void testSoftDeletes() throws Exception {
...
+    String normalizedTableName = "test_table";
+    String fieldName = "field-with-dash";
+    String normalizedFieldName = "field_with_dash";
+
+    try {
+      Schema schema = Schema.recordOf(tableName,
+          Schema.Field.of(fieldName, Schema.of(Schema.Type.INT)));
+      bigQuery.create(DatasetInfo.newBuilder(dataset).build());
+      DDLEvent createTable = DDLEvent.builder()
+          .setOperation(DDLOperation.Type.CREATE_TABLE)
```
yaojiefeng: Can you also check if other DDL operation passes? You can modify the existing tests, i.e, add a bad field name, to achieve that.

yaojiefeng (commented):
Minor comments, please make sure you sign the CLA, otherwise you won't be able to merge

src/main/java/io/cdap/delta/bigquery/Schemas.java
```diff
@@ -127,9 +127,9 @@ private static Field convertToBigQueryField(Schema.Field field) {
       if (bqType == null) {
         throw new IllegalArgumentException(
           String.format("Field '%s' is of type '%s', which is not supported in BigQuery.",
-                        name, logicalType.getToken()));
+              normalizedName, logicalType.getToken()));
```
albertshau: fix indentation.

Same comment everywhere else, please make sure your IDE indentation settings are consistent with the rest of the project

src/test/java/io/cdap/delta/bigquery/BigQueryEventConsumerTest.java
```diff
@@ -482,6 +482,46 @@ public void testSoftDeletes() throws Exception {
     }
   }
 
+  @Test
+  public void testBigQuerySchemaNormalization() throws Exception {
+    String bucketName = "bqtest-" + UUID.randomUUID().toString();
```
albertshau, line 487: this is an existing issue, but it would be better if the unit test creates one bucket in a test wide setup method annotated with `@BeforeClass`, and deletes the bucket in an `@AfterClass` annotated method. The reduces the chance that tests fail due to a GCP transient error, but more importantly makes the tests cleaner, as we don't need to wrap everything in a try-finally

albertshau: this can be done in a separate PR though, as it's unrelated to column normalization and would require passing in a MockContext that has a different app name for each test to prevent conflicts.

jingtaowu3: That's a good idea. Will address that in a separate PR

src/main/java/io/cdap/delta/bigquery/Schemas.java
jingtaowu3: done

src/test/java/io/cdap/delta/bigquery/BigQueryEventConsumerTest.java
jingtaowu3: space added.

yaojiefeng (approved):
LGTM, please squash commits before merging

PR #174, opened 2022-09-08
CDAP-19589 - Use sort keys for reliable ordering of DML events for unordered streams

sumitjnn:
- Added tests (will be adding few more scenarios)
- Validated there are no checkstyle errors

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -501,7 +501,7 @@ public class BigQueryEventConsumer implements EventConsumer {
     }
     primaryKeyStore.put(tableId, primaryKeys);
     context.putState(getTableStateKey(tableId),
-                     Bytes.toBytes(GSON.toJson(new BigQueryTableState(primaryKeys))));
+            Bytes.toBytes(GSON.toJson(new BigQueryTableState(primaryKeys, getSortKeys(tableId)))));
   }
 
   private List<String> getPrimaryKeys(TableId targetTableId) throws IOException, DeltaFailureException {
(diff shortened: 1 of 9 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryTableState.java
```diff
@@ -28,14 +33,31 @@ public class BigQueryTableState {
 
   private final List<String> primaryKeys;
 
+  @Nullable
+  private List<Schema.Type> sortKeys;
+
   public BigQueryTableState(List<String> primaryKeys) {
     this.primaryKeys = Collections.unmodifiableList(new ArrayList<>(primaryKeys));
```
seanzhougoogle: nit: call the other constructor and pass osrtKeys as empty list

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -493,15 +500,14 @@ private void updatePrimaryKeys(TableId tableId, List<String> primaryKeys) throws
       return;
     }
     primaryKeyStore.put(tableId, primaryKeys);
-    context.putState(String.format("bigquery-%s-%s", tableId.getDataset(), tableId.getTable()),
+    context.putState(getTableStateKey(tableId),
                      Bytes.toBytes(GSON.toJson(new BigQueryTableState(primaryKeys))));
```
seanzhougoogle: when primary key is updated, we don't preserve the old sortkey in the table state ?

sumitjnn: Fixed

src/main/java/io/cdap/delta/bigquery/Constants.java
```diff
@@ -26,6 +26,8 @@ public final class Constants {
   public static final String ROW_ID = "_row_id";
   public static final String OPERATION = "_op";
   public static final String BATCH_ID = "_batch_id";
+  public static final String SORT_KEYS = "_sort";
```
seanzhougoogle, line 29: nit: _sort -> _sort_keys

sumitjnn: I wanted to keep is the struct field as `_sort` and the actual nested key as `_key_0` (column - `_sort._key_0`), wdyt?

sumitjnn:
@sahusanket @seanzhougoogle Could you please review the PR or approve if there are no further issues?

src/main/java/io/cdap/delta/bigquery/Constants.java
seanzhougoogle: I see what you meant , sg.

seanzhougoogle (approved):

sahusanket (approved):

PR #176, opened 2022-09-19
PLUGIN-1388 : handle dup job submission

sahusanket:


src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -259,4 +263,22 @@ private static StructuredRecord normalize(StructuredRecord record) {
...
+   * Tries to submit a BQ job. If there is an Already Exists Exception, it will fetch the existing job
+   * @param bigquery
+   * @param jobInfo
+   * @return BQ Job
+   */
+  public static Job createBigQueryJob(BigQuery bigquery, JobInfo jobInfo) {
+    try {
+      return bigquery.create(jobInfo);
+    } catch (BigQueryException e) {
+      if (e.getCode() == 409 && e.getReason().equals("duplicate")) {
```
sumitjnn: Should we just check for error code here, checking on error message can be fragile

sahusanket: Agreed.
As of now using error code might be fine single only Duplicate is mapped to it. https://cloud.google.com/bigquery/docs/error-messages

But they might more `non duplicate` issues to this code. What do you think ?

sumitjnn: Looking at https://cloud.google.com/bigquery/docs/error-messages, it seems these are string error codes (instead of error message) to differentiate specific scenarios and hence it should be fine to compare.  Reason seems to be nullable in `BigQueryException`, can we check `"duplicate".equals(e.getReason())`?

sumitjnn (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
sahusanket: good catch. Thanks , will change it

PR #184, opened 2022-09-27
CDAP-19447- Reducing bigQuery API calls by fetching all tables at once

Aryan-Verma:
The issue was when a customer runs a big replication job from a database to BigQuery of the order that it replicates 600 tables with 10 tasks instance, they get an error message stating “Exceeded rate limits: too many api requests per user per method for this user_method.”

The solution provided is instead of hitting the API for each table to check if it exists or not, we can first fetch all tables that exist in the dataset. This will require a single call. We will store all those tables in a single set and check one by one if each table exists or not.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
@@ -25,6 +25,7 @@ import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobId;
 import com.google.cloud.bigquery.JobInfo;
 import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.Table;
 import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableResult;
 import com.google.common.collect.Iterables;
@@ -42,6 +43,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Pattern;
@@ -100,13 +102,23 @@ public final class BigQueryUtils {
   static long getMaximumExistingSequenceNumberPerBatch(Set<SourceTable> allTables, String project,
                                                        @Nullable String datasetName, BigQuery bigQuery,
                                                        EncryptionConfiguration encryptionConfiguration) {
+    SourceTable table0 = allTables.stream().findFirst().get();
+    Set<TableId> existingTableIDs = new HashSet<>();
+    String dataset = datasetName != null ? normalizeDatasetName(datasetName) :
+      normalizeDatasetName(table0.getDatabase());
+    if (bigQuery.getDataset(dataset) != null) {
+      for (Table table : bigQuery.listTables(dataset).iterateAll()) {
+        existingTableIDs.add(table.getTableId());
+      }
+    }
+
     StringBuilder builder = new StringBuilder();
     builder.append("SELECT MAX(max_sequence_num) FROM (");
     List<String> maxSequenceNumQueryPerTable = new ArrayList<>();
     for (SourceTable table : allTables) {
       TableId tableId = TableId.of(project, datasetName != null ? normalizeDatasetName(datasetName) :
         normalizeDatasetName(table.getDatabase()), normalizeTableName(table.getTable()));
-      if (bigQuery.getTable(tableId) != null) {
+      if (existingTableIDs.contains(tableId)) {
         maxSequenceNumQueryPerTable.add(String.format("SELECT MAX(_sequence_num) as max_sequence_num FROM %s",
                                                       wrapInBackTick(tableId.getDataset(), tableId.getTable())));
       }
--- a/src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
+++ b/src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
@@ -17,10 +17,13 @@
 
 package io.cdap.delta.bigquery;
 
+import com.google.api.gax.paging.Page;
 import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.PageImpl;
 import com.google.cloud.bigquery.BigQuery;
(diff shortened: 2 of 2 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -33,39 +33,29 @@
...
-  // Valid BigQuery table names can contain only Unicode characters in category L (letter), M (mark), N (number),
-  // Pc (connector, including underscore), Pd (dash), Zs (space).
-  // See here: https://cloud.google.com/bigquery/docs/tables#table_naming
-  private static final Pattern VALID_TABLE_NAME_REGEX = Pattern.compile("[\\p{L}\\p{M}\\p{N}\\p{Pc}\\p{Pd}\\p{Zs}]+");
-  private static final Pattern INVALID_TABLE_NAME_REGEX =
-          Pattern.compile("[^\\p{L}\\p{M}\\p{N}\\p{Pc}\\p{Pd}\\p{Zs}]+");
-  private static final String BIG_QUERY_DUPLICATE_ERROR = "duplicate";
+  // according to big query dataset and table naming convention, valid name should only contain letters (upper or
+  // lower case), numbers, and underscores
+  private static final String VALID_NAME_REGEX = "[\\w]+";
```
sumitjnn: Why was this change required? As per https://cloud.google.com/bigquery/docs/tables#table_naming, table names can contain unicode characters as well, I don't think `[\\w]+` will handle unicode characters except `[A-Za-z0-9_]`

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -150,7 +150,7 @@ private static long executeAggregateQuery(BigQuery bigQuery, String query, Encry
     }
     QueryJobConfiguration jobConfig = jobConfigBuilder.build();
     JobId jobId = JobId.of(UUID.randomUUID().toString());
-    Job queryJob = createBigQueryJob(bigQuery, JobInfo.newBuilder(jobConfig).setJobId(jobId).build());
+    Job queryJob = bigQuery.create(JobInfo.newBuilder(jobConfig).setJobId(jobId).build());
```
sumitjnn: Why this change?

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -264,22 +249,4 @@ private static StructuredRecord normalize(StructuredRecord record) {
...
     return BACKTICK + datasetName + "." + tableName + BACKTICK;
   }
-
-  /**
-   * Tries to submit a BQ job. If there is an Already Exists Exception, it will fetch the existing job
-   * @param bigquery
-   * @param jobInfo
-   * @return BQ Job
-   */
-  public static Job createBigQueryJob(BigQuery bigquery, JobInfo jobInfo) {
```
sumitjnn: The method is being used in `BigQueryEventConsumer`, why did we remove this?

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -198,20 +186,17 @@ public static String normalizeTableName(String name) {
...
       return name;
     }
 
     // replace invalid chars with underscores if there are any
-    if (useExtendedCharset && !VALID_TABLE_NAME_REGEX.matcher(name).matches()) {
-      name = INVALID_TABLE_NAME_REGEX.matcher(name).replaceAll("_");
-    }
-    if (!useExtendedCharset && !VALID_DATASET_NAME_REGEX.matcher(name).matches()) {
-      name = INVALID_DATASET_NAME_REGEX.matcher(name).replaceAll("_");
+    if (!name.matches(VALID_NAME_REGEX)) {
```
sumitjnn: Using a `Pattern` is more efficient here, as otherwise `String.matches()` internally will create and compile a new `Pattern` object internally every time, please refer to https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#matches-java.lang.String-java.lang.CharSequence-

sumitjnn (changes requested):

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
```diff
@@ -155,6 +163,7 @@ public void testNormalizeFieldName() {
     public void testGetMaximumExistingSequenceNumberZeroInvocations() throws Exception {
       // Zero Tables
       Set<SourceTable> allTables = generateSourceTableSet(0);
+      Mockito.when(bigQueryMock.listTables((String) null)).thenReturn(generateBQTablesPage(0));
```
sahusanket: use `ArgumentMatchers.anyString()`

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
```diff
@@ -350,4 +365,45 @@ private static Set<SourceTable> generateSourceTableSet(int noOfTables) {
...
+      @Override
+      public Iterable<Table> iterateAll() {
+        List<Table> tableList = new ArrayList<>();
+
+        for (int i = 1; i <= num; i++) {
+          // Create Table
+          Table tableMock2 = Mockito.mock(Table.class);
+
+          String tableName = TABLE_PREFIX + i;
+          TableId tableId = TableId.of("testproject", DATASET, tableName);
```
sahusanket: instead of using hardcoded "testproject" , 
Declare a `private static final String PROJECT = "testproject";` just like `DATASET` and use it .

sahusanket (changes requested):

sumitjnn (approved):
Please ensure there are no checkstyle/unit tests errors before merging with `mvn clean install  -Drat.ignoreErrors=true`

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
Aryan-Verma: Okay

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
Aryan-Verma: Okay

sahusanket (approved):

PR #190, opened 2022-10-21
CDAP-19939 Fix handling of high precision datetime

sumitjnn:
Inserting datetime value with precision higher than microseconds fails in BigQuery as it supports only upto microsecond precision https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime_type. 
This PR fixes the issue by trimming extra precision digits from the value when writing the record.
Note that we already show a warning to user about loss of precision during replication pipeline assessment.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/StructuredRecordToJson.java
+++ b/src/main/java/io/cdap/delta/bigquery/StructuredRecordToJson.java
@@ -62,7 +62,6 @@ public final class StructuredRecordToJson {
    */
   private static final Pattern LOGICAL_DATE_PATTERN =
     Pattern.compile("\\d{4}-\\d{1,2}-\\d{1,2}([ T]\\d{1,2}:\\d{1,2}:\\d{1,2}(\\.(\\d+))?)?");
-  private static final int LOGICAL_DATE_WITH_TIMEFRACTION_GROUP_COUNT = 3;
   private static final int TIME_FRACTION_GROUP = 3;
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/StructuredRecordToJson.java
```diff
@@ -185,6 +207,31 @@ private static void writeSimpleTypes(JsonWriter writer, String name, boolean isA
     }
   }
 
+  private static String checkAndTrimToMaxSupportedPrecision(String strValue) {
+    Matcher matcher = LOGICAL_DATE_PATTERN.matcher(strValue);
+    if (matcher.matches()) {
+      if (matcher.groupCount() == LOGICAL_DATE_WITH_TIMEFRACTION_GROUP_COUNT) {
```
sahusanket: nit: merge into single if ?

sumitjnn: I think we can just remove the groupCount check it seems to only depend on the pattern and we already have null check on L216

src/main/java/io/cdap/delta/bigquery/StructuredRecordToJson.java
```diff
@@ -185,6 +206,29 @@ private static void writeSimpleTypes(JsonWriter writer, String name, boolean isA
     }
   }
 
+  private static String checkAndTrimToMaxSupportedPrecision(String strValue) {
+    Matcher matcher = LOGICAL_DATE_PATTERN.matcher(strValue);
+    if (matcher.matches()) {
+      String timeFraction = matcher.group(TIME_FRACTION_GROUP);
+      //matcher.group returns null for a group if an optional group did not exist in the string
+      if (timeFraction != null) {
+        if (timeFraction.length() > MAX_LOGICAL_DATE_TIME_FRACTION_PRECISION) {
```
sahusanket: Similar nit :  `timeFraction != null && timeFraction.length() > MAX_LOGICAL_DATE_TIME_FRACTION_PRECISION`

single if ?

sahusanket (approved):

PR #196, opened 2022-11-25
Log more error list when operation fails (CDAP-19423)

yjhawar:
When the BigQuery job operation fails, the error list is not logged. This change improves logging by displaying the error list.(CDAP-19423)

image

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -1277,7 +1277,7 @@ public class BigQueryEventConsumer implements EventConsumer {
   }
 
   private void runWithRetries(ContextualRunnable runnable, long retryDelay, String dataset, String schema, String table,
-                              String onFailedAttemptMessage, String retriesExhaustedMessage) throws DeltaFailureException, InterruptedException {
+    String onFailedAttemptMessage, String retriesExhaustedMessage) throws DeltaFailureException, InterruptedException {
     RetryPolicy<Object> retryPolicy = createBaseRetryPolicy(retryDelay)
       .onFailedAttempt(failureContext -> {
         Throwable t = failureContext.getLastFailure();
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -1277,16 +1277,16 @@ public class BigQueryEventConsumer implements EventConsumer {
   }
 
   private void runWithRetries(ContextualRunnable runnable, long retryDelay, String dataset, String schema, String table,
-    String onFailedAttemptMessage, String retriesExhaustedMessage) throws DeltaFailureException, InterruptedException {
+                              String onFailedAttemptMessage, String retriesExhaustedMessage) throws DeltaFailureException, InterruptedException {
     RetryPolicy<Object> retryPolicy = createBaseRetryPolicy(retryDelay)
       .onFailedAttempt(failureContext -> {
         Throwable t = failureContext.getLastFailure();
         LOG.error(onFailedAttemptMessage, t);
         //Logging error list
-        List<BigQueryError> errors = ((BigQueryException) t).getErrors();
-        if (errors.size() > 1) {
-          for (int i = 1; i < errors.size(); i++) {
-            LOG.error(errors.get(i).getMessage());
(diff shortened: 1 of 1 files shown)


sumitjnn:
@yjhawar PR contains lot of formatting changes, can you confirm if you have setup IDE as per https://cdap.atlassian.net/wiki/spaces/CE/pages/1594098619/Coding+Standard

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -1282,6 +1282,13 @@ private void runWithRetries(ContextualRunnable runnable, long retryDelay, String
       .onFailedAttempt(failureContext -> {
         Throwable t = failureContext.getLastFailure();
         LOG.error(onFailedAttemptMessage, t);
+        //Logging error list
+        List<BigQueryError> errors = ((BigQueryException) t).getErrors();
+        if (errors.size() > 1) {
```
sumitjnn: 1. Check errors is not null
2. Why are we not logging if we have a single error (errors.size() == 1) ?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -1282,6 +1282,13 @@ private void runWithRetries(ContextualRunnable runnable, long retryDelay, String
...
+          for (int i = 1; i < errors.size(); i++) {
```
sumitjnn: Loop should start from 0

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -1282,6 +1282,13 @@ private void runWithRetries(ContextualRunnable runnable, long retryDelay, String
       .onFailedAttempt(failureContext -> {
         Throwable t = failureContext.getLastFailure();
         LOG.error(onFailedAttemptMessage, t);
+        //Logging error list
+        List<BigQueryError> errors = ((BigQueryException) t).getErrors();
```
sumitjnn: I don't think we can assume that exception type will always be BigQueryException. 

This should be guarded by  t instance BigQueryException

sumitjnn (changes requested):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
yjhawar: 2. In the errors list ,the first msg is "Error while reading data, error message: JSON table encountered too many errors, giving up. Rows: 1; errors: 1. Please look into the errors[] collection for more details." which is logged with the throwable

sumitjnn: That's an implementation detail of BigQuery client which we should not depend upon as it can change in future, I would suggest to log the complete error list even if first item in the list happens to be same as the error message of the exception.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -1277,11 +1277,18 @@ static String createMergeQuery(TableId targetTableId, List<String> primaryKeys,
...
+                              String onFailedAttemptMessage, String retriesExhaustedMessage) throws DeltaFailureException, InterruptedException {
     RetryPolicy<Object> retryPolicy = createBaseRetryPolicy(retryDelay)
       .onFailedAttempt(failureContext -> {
         Throwable t = failureContext.getLastFailure();
         LOG.error(onFailedAttemptMessage, t);
+        //Logging error list
+        if (t instanceof BigQueryException) {
+          List<BigQueryError> errors = ((BigQueryException) t).getErrors();
+          if (errors != null) {
+            errors.forEach(err -> LOG.error(err.getMessage()));
```
sumitjnn, line 1289: BigQueryError has few other properties than message - reason, location
Would it make sense to log those too?

sumitjnn:
@yjhawar Please fix build failures

sumitjnn (approved):

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
yjhawar: It might not always be useful and relevant.This info added in such cases creates confusion.Error message is pretty much clear to take action.

PR #199, opened 2022-12-29
CDAP-20214: Fix IllegalArgumentException: Provided dataset is null or empty

sumitjnn:
https://cdap.atlassian.net/browse/CDAP-20214

Today if target dataset is null the plugin uses source database name as target dataset name. But we are missing the handling when target dataset is empty string, which results in pipeline failure.

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
@@ -34,6 +34,7 @@ import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.delta.api.DMLEvent;
 import io.cdap.delta.api.SourceTable;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -179,6 +179,21 @@ private static long executeAggregateQuery(BigQuery bigQuery, String query, Encry
...
+  /**
+   * Get normalized dataset name for target dataset
+   * Method returns normalized datasetName if it is not empty
+   * Otherwise it returns normalized form of databaseName
+   * @param datasetName dataset name to be normalized
+   * @param databaseName database name to be used if datasetName is empty
+   * @return normalized datasetName if it is not empty otherwise normalized form of databaseName
+   */
+  public static String getNormalizedDatasetName(@Nullable String datasetName, String databaseName) {
+    if (StringUtils.isNotEmpty(datasetName)) {
```
itsankit-google: we can use `Strings.IsNullOrEmtpy()` method from guava instead of introducing new dependency for this case.

sumitjnn: Thanks, removed

itsankit-google (approved):
lgtm

PR #201, opened 2023-01-13
CDAP-20276: Fix data inconsistency issue in case of retries in BigQuery replication plugin

sumitjnn:
### Issue

The plugin runs following BQ jobs as part of replication pipeline

1. Load job to insert snapshot events directly to the target table
2. Load job to insert CDC events directly to the staging table
3. Query job to merge CDC events from staging to target table

In case of transient network errors just after the create job request reaches BQ, the create or check status calls for job can fail and a retry would be triggered after the retry interval (90 sec). Retry would run a new job which would lead to duplicate data in staging/target table. Duplicate data in staging table can intern lead to duplicate data in target table or pipeline failure due to error `UPDATE/MERGE must match at most one source row for each target row`

### Fix
In the retry attempt, check if there was any non failing job from the previous attempts, and only create a new job if there was none.

### Testing

- Validated failure scenarios in sandbox
- Unit tests

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -164,9 +164,6 @@ public class BigQueryEventConsumer implements EventConsumer {
   private static final Gson GSON = new Gson();
   private static final String RETAIN_STAGING_TABLE = "retain.staging.table";
   private static final String DIRECT_LOADING_IN_PROGRESS_PREFIX = "bigquery-direct-load-in-progress-";
-  public static final String MERGE_JOB_TYPE = "merge";
-  public static final String LOAD_STAGING_JOB_TYPE = "stage";
-  public static final String LOAD_TARGET_JOB_TYPE = "load";
 
   private final DeltaTargetContext context;
   private final BigQuery bigQuery;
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -163,6 +164,9 @@ public class BigQueryEventConsumer implements EventConsumer {
   private static final Gson GSON = new Gson();
   private static final String RETAIN_STAGING_TABLE = "retain.staging.table";
   private static final String DIRECT_LOADING_IN_PROGRESS_PREFIX = "bigquery-direct-load-in-progress-";
+  public static final String MERGE_JOB_TYPE = "merge";
```
sahusanket: Should we convert these 3 to an ENUM  ?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -781,27 +829,56 @@ private void loadTable(TableId tableId, TableBlob blob, boolean directLoadToTarg
...
+  }
+
+  private void mergeStagingTable(TableId stagingTableId, TableBlob blob,
+                                 int attemptNumber) throws InterruptedException, IOException, DeltaFailureException {
+
+    LOG.info("Merging batch {} for {}.{} {}", blob.getBatchId(), blob.getDataset(), blob.getTable(),
+             attemptNumber > 0 ? "attempt: " + attemptNumber : "");
+
+    Job mergeJob = null;
+    if (attemptNumber > 0) {
```
sahusanket, line 863: can we refactor this to a separate function and reuse here and above ( line 747) as well ?

sahusanket (approved):
Minor NITs, otherwise LGTM

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sumitjnn: Done

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sumitjnn: Done

PR #207, opened 2023-01-30
Add plugin level retry when writing event to GCS as part of DML event processing

yjhawar:
- Added unit tests

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -193,6 +193,11 @@ public class BigQueryEventConsumer implements EventConsumer {
   private long latestSequenceNum;
   private Exception flushException;
   private final AtomicBoolean shouldStop;
+  private RetryPolicy<Object> gcsWriterRetryPolicy = new RetryPolicy<>()
+                                                .withMaxAttempts(25)
+                                                .withMaxDuration(Duration.of(2, ChronoUnit.MINUTES))
+                                                .withBackoff(1, 30, ChronoUnit.SECONDS)
+                                                .withJitter(0.1);
 
   private enum JobType {
     LOAD_STAGING("stage", false), LOAD_TARGET("load", true), MERGE_TARGET("merge", true);
@@ -593,7 +598,8 @@ public class BigQueryEventConsumer implements EventConsumer {
       .setTableName(normalizedTableName)
       .build();
     long sequenceNumber = sequencedEvent.getSequenceNumber();
-    gcsWriter.write(new Sequenced<>(normalizedDMLEvent, sequenceNumber));
+    Failsafe.with(gcsWriterRetryPolicy)
+            .run(() -> gcsWriter.write(new Sequenced<>(normalizedDMLEvent, sequenceNumber)));
 
     TableId tableId = TableId.of(project, normalizedDatabaseName, normalizedTableName);
 
--- a/src/test/java/io/cdap/delta/bigquery/BigQueryConsumerTest.java
+++ b/src/test/java/io/cdap/delta/bigquery/BigQueryConsumerTest.java
@@ -46,7 +46,9 @@ import io.cdap.delta.api.Offset;
 import io.cdap.delta.api.Sequenced;
 import org.apache.avro.file.DataFileWriter;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentMatchers;
 import org.mockito.Mock;
@@ -115,6 +117,8 @@ public class BigQueryConsumerTest {
   private Job job;
   @Mock
   private DataFileWriter dataFileWriter;
+  @Rule
+  private ExpectedException exceptionRule = ExpectedException.none();
 
   @Before
   public void setup() throws Exception {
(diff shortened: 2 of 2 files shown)


sumitjnn:
@yjhawar I think there should not be any failures due to in memory writes, but only when writes are pushed to GCS from the buffer, please fix the PR title

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -593,7 +593,25 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
       .setTableName(normalizedTableName)
       .build();
     long sequenceNumber = sequencedEvent.getSequenceNumber();
-    gcsWriter.write(new Sequenced<>(normalizedDMLEvent, sequenceNumber));
+    try {
+      Failsafe.with(new RetryPolicy<>()
```
sumitjnn: to avoid creating retrypolicy object every time, can we create the retry policy object as a field in BigQueryEventConsumer, maybe call it gcsWriterRetryPolicy ?

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -593,7 +593,25 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
...
+                      .withMaxDuration(Duration.of(2, ChronoUnit.MINUTES))
+                      .withBackoff(1, 30, ChronoUnit.SECONDS)
+                      .withJitter(0.1))
+              .run(() -> gcsWriter.write(new Sequenced<>(normalizedDMLEvent, sequenceNumber)));
+    } catch (TimeoutExceededException e) {
+      // if the retry timeout was reached, throw a DeltaFailureException to fail the pipeline immediately
+      DeltaFailureException exc = new DeltaFailureException
+              (String.format("Unable to write event %s to bytes.", event), e);
+      flushException = exc;
+      throw exc;
```
sumitjnn: Let skip catching TimeoutExceededException, as we don't want to fail the pipeline permanently but the let worker retry

src/test/java/io/cdap/delta/bigquery/BigQueryConsumerTest.java
```diff
@@ -126,6 +128,7 @@ public void setup() throws Exception {
     Mockito.when(dataFileWriter.create(Mockito.any(org.apache.avro.Schema.class), Mockito.any(OutputStream.class)))
       .thenReturn(dataFileWriter);
     PowerMockito.whenNew(DataFileWriter.class).withAnyArguments().thenReturn(dataFileWriter);
+    PowerMockito.whenNew(MultiGCSWriter.class).withAnyArguments().thenReturn(gcsWriter);
```
sumitjnn: Instead of this can we just mock `dataFileWriter.append` method to avoid using powermock for MultiGCSWriter?

src/test/java/io/cdap/delta/bigquery/BigQueryConsumerTest.java
```diff
@@ -305,6 +308,37 @@ public void testConsumerCommitFailureRetries() throws Exception {
...
+                                                                    LOAD_INTERVAL_ONE_SECOND, "_staging",
+                                                                    false, null, 2L,
+                                                                    DATASET, false);
+
+    try {
+      eventConsumer.applyDML(new Sequenced<>(insert1Event, 0));
+    } catch (IllegalStateException e) {
+      //Verify that retry happens
+      Mockito.verify(gcsWriter, Mockito.atLeast(2)).write(Mockito.any(Sequenced.class));
+    }
```
sumitjnn:
```suggestion
    try {
      exceptionRule.expect(IllegalStateException.class);
      eventConsumer.applyDML(new Sequenced<>(insert1Event, 0));
    } finally {
      //Verify that retry happens
      Mockito.verify(gcsWriter, Mockito.atLeast(2)).write(Mockito.any(Sequenced.class));
      eventConsumer.stop();
    }
```

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
yjhawar: Have updated the code.Both timeout and interrupted exception handling isn't required.

src/test/java/io/cdap/delta/bigquery/BigQueryConsumerTest.java
yjhawar: I didn't get this. gcsWriter.write operation is performed which has to be mocked.How can it be replaced by this.

sumitjnn: It can be replaced because if you trace the call chain eventually it will reach to `dataFileWriter.append`

src/test/java/io/cdap/delta/bigquery/BigQueryConsumerTest.java
```diff
@@ -126,6 +132,7 @@ public void setup() throws Exception {
     Mockito.when(dataFileWriter.create(Mockito.any(org.apache.avro.Schema.class), Mockito.any(OutputStream.class)))
       .thenReturn(dataFileWriter);
     PowerMockito.whenNew(DataFileWriter.class).withAnyArguments().thenReturn(dataFileWriter);
+    //PowerMockito.whenNew(MultiGCSWriter.class).withAnyArguments().thenReturn(gcsWriter);
```
sumitjnn: Nit: Remove this line

sumitjnn (approved):
lgtm with minor nit

PR #209, opened 2023-02-01
Add retry for BQ calls during plugin initialization 

yjhawar:
- Added retry for BQ call to retrieve maximum existing sequence number and GCS calls to get/create staging buckets.
- Added unit tests

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
@@ -77,11 +77,18 @@ public class BigQueryTarget implements DeltaTarget {
   private static final Set<Integer> RATE_LIMIT_EXCEEDED_CODES = new HashSet<>(Arrays.asList(400, 403));
   private  static final int BILLING_TIER_LIMIT_EXCEEDED_CODE = 400;
   private  static final String BILLING_TIER_LIMIT_EXCEEDED_REASON = "billingTierLimitExceeded";
+  private  static final int RETRY_COUNT = 25;
+  private final int retryCount;
   private final Conf conf;
 
   @SuppressWarnings("unused")
   public BigQueryTarget(Conf conf) {
+    this(conf, RETRY_COUNT);
+  }
+  @SuppressWarnings("unused")
+  public BigQueryTarget(Conf conf, int retryCount) {
     this.conf = conf;
+    this.retryCount = retryCount;
   }
 
   @Override
@@ -106,16 +113,19 @@ public class BigQueryTarget implements DeltaTarget {
       .getService();
 
     RetryPolicy<Object> retryPolicy = createBaseRetryPolicy()
-            .handleIf(throwable -> {
-              if (throwable instanceof BigQueryException) {
-                BigQueryException t = (BigQueryException) throwable;
+            .handleIf(ex -> {
+              if (ex.getCause() instanceof IOException) {
+                return true;
+              }
+              if (ex instanceof BigQueryException) {
+                BigQueryException t = (BigQueryException) ex;
                 int code = t.getCode();
                 String reason = t.getError() != null ? t.getError().getReason() : null;
                 boolean isRateLimitExceeded =  RATE_LIMIT_EXCEEDED_CODES.contains(code)
                         && RATE_LIMIT_EXCEEDED_REASON.equals(reason);
                 boolean isBillingTierLimitExceeded = code == BILLING_TIER_LIMIT_EXCEEDED_CODE
                         && BILLING_TIER_LIMIT_EXCEEDED_REASON.equals(reason);
-                return (t.isRetryable() || isRateLimitExceeded || isBillingTierLimitExceeded);
+                return t.isRetryable() || isRateLimitExceeded || isBillingTierLimitExceeded;
               }
               return false;
             });
@@ -179,7 +189,7 @@ public class BigQueryTarget implements DeltaTarget {
           // bucket is created by another worker instance after this worker instance
           // determined that the bucket does not exists. Ignore error if bucket already exists.
           if (e.getCode() != CONFLICT) {
-            throw new IOException(
+            throw new RuntimeException(
                     String.format("Unable to create staging bucket '%s' in project '%s'. " +
                             "Please make sure the service account has permission to create buckets, " +
                             "or create the bucket before starting the program.", stagingBucketName, project), e);
@@ -220,7 +230,7 @@ public class BigQueryTarget implements DeltaTarget {
 
   private <T> RetryPolicy<T> createBaseRetryPolicy() {
     RetryPolicy<T> retryPolicy = new RetryPolicy<>();
-    return retryPolicy.withMaxAttempts(2)
+    return retryPolicy.withMaxAttempts(retryCount)
             .withMaxDuration(Duration.of(2, ChronoUnit.MINUTES))
             .withBackoff(1, 30, ChronoUnit.SECONDS)
             .withJitter(0.1);
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
@@ -20,6 +20,7 @@ import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.ServiceOptions;
 import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.BigQueryOptions;
 import com.google.cloud.bigquery.EncryptionConfiguration;
 import com.google.cloud.storage.Bucket;
@@ -41,6 +42,8 @@ import io.cdap.delta.api.DeltaTargetContext;
 import io.cdap.delta.api.EventConsumer;
 import io.cdap.delta.api.assessment.StandardizedTableDetail;
 import io.cdap.delta.api.assessment.TableAssessor;
+import net.jodah.failsafe.Failsafe;
+import net.jodah.failsafe.RetryPolicy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,7 +51,12 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
@@ -65,6 +73,10 @@ public class BigQueryTarget implements DeltaTarget {
   private static final String GCS_SCHEME = "gs://";
   private static final String GCP_CMEK_KEY_NAME = "gcp.cmek.key.name";
   private static final int MAX_TABLES_PER_QUERY = 1000;
+  private  static final String RATE_LIMIT_EXCEEDED_REASON = "rateLimitExceeded";
+  private static final Set<Integer> RATE_LIMIT_EXCEEDED_CODES = new HashSet<>(Arrays.asList(400, 403));
+  private  static final int BILLING_TIER_LIMIT_EXCEEDED_CODE = 400;
+  private  static final String BILLING_TIER_LIMIT_EXCEEDED_REASON = "billingTierLimitExceeded";
   private final Conf conf;
 
   @SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -178,6 +201,27 @@ private static String stringifyPipelineId(DeltaPipelineId pipelineId) {
...
+    retryPolicy.abortOn(throwable -> {
+      if (throwable instanceof BigQueryException) {
+        BigQueryException t = (BigQueryException) throwable;
+        int code = t.getCode();
+        String reason = t.getError() != null ? t.getError().getReason() : null;
+        boolean isRateLimitExceeded =  RATE_LIMIT_EXCEEDED_CODES.contains(code)
+                && RATE_LIMIT_EXCEEDED_REASON.equals(reason);
+        boolean isBillingTierLimitExceeded = code == BILLING_TIER_LIMIT_EXCEEDED_CODE
+                && BILLING_TIER_LIMIT_EXCEEDED_REASON.equals(reason);
+        return !(t.isRetryable() || isRateLimitExceeded || isBillingTierLimitExceeded);
```
sumitjnn: Nit: To avoid the negation we can use handleIf instead of abortOn to only retry if method returns true

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -178,6 +201,27 @@ private static String stringifyPipelineId(DeltaPipelineId pipelineId) {
                                pipelineId.getGeneration());
   }
 
+  private <T> RetryPolicy<T> createRetryPolicyForBq() {
+    RetryPolicy<T> retryPolicy = new RetryPolicy<>();
+    retryPolicy.abortOn(throwable -> {
+      if (throwable instanceof BigQueryException) {
```
sumitjnn: I think we also wanted to retry if throwable or cause of BigQueryException is instance of IOException

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -125,30 +137,41 @@ public EventConsumer createConsumer(DeltaTargetContext context) throws IOExcepti
...
-        BucketInfo.Builder builder = BucketInfo.newBuilder(stagingBucketName);
-        if (cmekKey != null) {
-          builder.setDefaultKmsKeyName(cmekKey);
+    RetryPolicy<Object> retryPolicy = new RetryPolicy<>()
+            .withMaxAttempts(25)
+            .withMaxDuration(Duration.of(2, ChronoUnit.MINUTES))
+            .withBackoff(1, 30, ChronoUnit.SECONDS)
+            .withJitter(0.1)
+            .abortOn(throwable -> !((throwable instanceof IOException) ||
+                    (throwable instanceof StorageException && ((StorageException) throwable).isRetryable())));
```
sumitjnn:
```suggestion
            .handleIf(ex -> ex instanceof IOException 
            || ex.getCause() instanceof IOException         
            || (ex instanceof StorageException && ((StorageException) ex).isRetryable()));
```

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
sumitjnn: Would this ever match if `BigQueryUtils.getMaximumExistingSequenceNumber` always throws RuntimeException ?

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -125,30 +137,41 @@ public EventConsumer createConsumer(DeltaTargetContext context) throws IOExcepti
...
-    if (bucket == null) {
-      try {
-        BucketInfo.Builder builder = BucketInfo.newBuilder(stagingBucketName);
-        if (cmekKey != null) {
-          builder.setDefaultKmsKeyName(cmekKey);
+    RetryPolicy<Object> retryPolicy = new RetryPolicy<>()
+            .withMaxAttempts(25)
+            .withMaxDuration(Duration.of(2, ChronoUnit.MINUTES))
+            .withBackoff(1, 30, ChronoUnit.SECONDS)
+            .withJitter(0.1)
```
sumitjnn: Can we create base retry policy with common settings and configure abortOn/handleIf separately for BQ/GCS?

src/test/java/io/cdap/delta/bigquery/BigQueryTargetTest.java
```diff
@@ -41,4 +113,122 @@ public void testStagingBucketName() {
...
+            new BigQueryError(BILLING_TIER_LIMIT_EXCEEDED_REASON, null, null)));
+    exceptions.add(new BigQueryException(RATE_LIMIT_EXCEEDED_CODES.stream().findAny().get(), null,
+            new BigQueryError(RATE_LIMIT_EXCEEDED_REASON, null, null)));
+
+    BigQueryTarget bqTarget = new BigQueryTarget(conf);
+
+    for (Throwable exception: exceptions) {
+      PowerMockito.mockStatic(BigQueryUtils.class);
+      PowerMockito.doThrow(exception).when(BigQueryUtils.class);
+      BigQueryUtils.getMaximumExistingSequenceNumber(Mockito.anySet(), Mockito.anyString(),
```
sumitjnn: To simulate real behaviour (RuntimeException thrown from the method), it would be better to not mock `BigQueryUtils.getMaximumExistingSequenceNumber` itself, but instead the internal BigQuery call, please refer `BigQueryConsumerTest.testPermanentFailureIsNotRetriedInProcessDDL` in latest code.

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
yjhawar: This method doesn't throw IOException. It throws Runtime exception.https://github.com/data-integrations/bigquery-delta-plugins/blob/62f2cc7968924818478370f4190bb876be648340/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java#L107

yjhawar: We never retry runtime exception.

sumitjnn: The point was that getMaximumExistingSequenceNumber would convert BigQueryException to RuntimeException, so this check would not work

sumitjnn: In that case what's the point of checking BigQueryException?

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -136,10 +136,12 @@ static long getMaximumExistingSequenceNumberPerBatch(Set<SourceTable> allTables,
...
                                                                                               builder.toString(),
                                                                                               encryptionConfiguration);
-    } catch (Exception e) {
+    } catch (InterruptedException e) {
       throw new RuntimeException("Failed to compute the maximum sequence number among all the target tables " +
-                                   "selected for replication. Please make sure that if target tables exists, " +
-                                   "they should have '_sequence_num' column in them.", e);
+              "selected for replication. Please make sure that if target tables exists, " +
+              "they should have '_sequence_num' column in them.", e);
+    } catch (Exception e) {
```
sumitjnn: There is no need to catch these exceptions and then rethrow them. It would be better to remove both the catch blocks for InterruptedException and Exception, instead add throws InterruptedException to method signatures of `getMaximumExistingSequenceNumber` and `getMaximumExistingSequenceNumberPerBatch`

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -93,13 +105,33 @@ public void initialize(DeltaTargetContext context) throws Exception {
...
+            .handleIf(throwable -> {
+              if (throwable instanceof BigQueryException) {
+                BigQueryException t = (BigQueryException) throwable;
+                int code = t.getCode();
+                String reason = t.getError() != null ? t.getError().getReason() : null;
+                boolean isRateLimitExceeded =  RATE_LIMIT_EXCEEDED_CODES.contains(code)
+                        && RATE_LIMIT_EXCEEDED_REASON.equals(reason);
+                boolean isBillingTierLimitExceeded = code == BILLING_TIER_LIMIT_EXCEEDED_CODE
+                        && BILLING_TIER_LIMIT_EXCEEDED_REASON.equals(reason);
+                return (t.isRetryable() || isRateLimitExceeded || isBillingTierLimitExceeded);
```
sumitjnn: nit: remove parenthesis
```suggestion
                return t.isRetryable() || isRateLimitExceeded || isBillingTierLimitExceeded;
```

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -93,13 +105,33 @@ public void initialize(DeltaTargetContext context) throws Exception {
       .build()
       .getService();
 
-    long maximumExistingSequenceNumber =
-      BigQueryUtils.getMaximumExistingSequenceNumber(context.getAllTables(), project, conf.getDatasetName(),
-                                                     bigQuery, encryptionConfig, MAX_TABLES_PER_QUERY);
+    RetryPolicy<Object> retryPolicy = createBaseRetryPolicy()
+            .handleIf(throwable -> {
+              if (throwable instanceof BigQueryException) {
```
sumitjnn: Also enable retry for IOException

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -125,30 +157,38 @@ public EventConsumer createConsumer(DeltaTargetContext context) throws IOExcepti
...
+          if (conf.stagingBucketLocation != null && !conf.stagingBucketLocation.trim().isEmpty()) {
+            builder.setLocation(conf.stagingBucketLocation);
+          }
+          b = storage.create(builder.build());
+        } catch (StorageException e) {
+          // It is possible that in multiple worker instances scenario
+          // bucket is created by another worker instance after this worker instance
+          // determined that the bucket does not exists. Ignore error if bucket already exists.
+          if (e.getCode() != CONFLICT) {
+            throw new IOException(
```
sumitjnn: Instead of IOException, let's throw RuntimeException here as otherwise it will be retried due to the retry policy

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -178,6 +218,14 @@ private static String stringifyPipelineId(DeltaPipelineId pipelineId) {
                                pipelineId.getGeneration());
   }
 
+  private <T> RetryPolicy<T> createBaseRetryPolicy() {
+    RetryPolicy<T> retryPolicy = new RetryPolicy<>();
+    return retryPolicy.withMaxAttempts(2)
```
sumitjnn: Change max attempts to 25

src/test/java/io/cdap/delta/bigquery/BigQueryTargetTest.java
```diff
@@ -41,4 +170,117 @@ public void testStagingBucketName() {
...
+      Mockito.verify(storage, Mockito.atLeast(2)).get(Mockito.nullable(String.class));
+    }
+  }
+
+  @Test
+  public void testCreateConsumerNonRetryableFailure() throws Exception {
+    Throwable exception = new StorageException(501, null);
+    Mockito.when(storage.get(Mockito.anyString())).thenThrow(exception);
+
+    BigQueryTarget bqTarget = new BigQueryTarget(conf);
```
sumitjnn: Create another constructor in BigQueryTarget which takes max retry attempts to avoid blocking the test for minutes

src/test/java/io/cdap/delta/bigquery/BigQueryTargetTest.java
```diff
@@ -16,21 +16,150 @@
...
+  @Mock
+  private BigQuery bigQuery;
+  @Mock
+  private BucketInfo bucketInfo;
+  @Mock
+  private Storage storage;
+  @Mock
+  StorageOptions storageOptions;
+  @Mock
+  BigQueryOptions bigqueryOptions;
```
sumitjnn: Nit: make these private too

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
yjhawar: since it doesn't directly throw IOException, will check for cause.

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
```diff
@@ -65,11 +73,21 @@ public class BigQueryTarget implements DeltaTarget {
...
+  private  static final String RATE_LIMIT_EXCEEDED_REASON = "rateLimitExceeded";
+  private static final Set<Integer> RATE_LIMIT_EXCEEDED_CODES = new HashSet<>(Arrays.asList(400, 403));
+  private  static final int BILLING_TIER_LIMIT_EXCEEDED_CODE = 400;
+  private  static final String BILLING_TIER_LIMIT_EXCEEDED_REASON = "billingTierLimitExceeded";
+  private final int retryCount;
   private final Conf conf;
 
   @SuppressWarnings("unused")
   public BigQueryTarget(Conf conf) {
+    this(conf, 25);
```
sumitjnn: Create constant for default value of retry count

src/main/java/io/cdap/delta/bigquery/BigQueryTarget.java
yjhawar: here, in this case it will not retry for any storage exception.We should throw e and add try-catch with Runtime for failsafe

sumitjnn (approved):
LGTM

PR #213, opened 2023-03-16
CDAP-19929 - Add logging for critical events at default log level

yjhawar:
- Metadata for DDL events logged at INFO level
- Schema object logged at Debug level

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -343,7 +343,10 @@ public class BigQueryEventConsumer implements EventConsumer {
   private void handleDDL(DDLEvent event, String normalizedDatabaseName, String normalizedTableName,
                          String normalizedStagingTableName)
     throws IOException, DeltaFailureException, InterruptedException {
-
+    LOG.info("DDL Event={}, sequenceNumber={}", event, latestSequenceNum);
+    if (event.getSchema() != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -343,7 +343,10 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
   private void handleDDL(DDLEvent event, String normalizedDatabaseName, String normalizedTableName,
                          String normalizedStagingTableName)
     throws IOException, DeltaFailureException, InterruptedException {
-
+    LOG.info("DDL Event={}, sequenceNumber={}", event, latestSequenceNum);
```
sumitjnn: Sequence number is not incremented for DDL events hence should be skipped

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -343,7 +343,10 @@ public synchronized void applyDDL(Sequenced<DDLEvent> sequencedEvent) throws Exc
...
+    if (event.getSchema() != null) {
```
sumitjnn: Debug logging should be guarded by LOG.isDebugEnabled()

```suggestion
    if (LOG.isDebugEnabled() && event.getSchema() != null) {
```

sumitjnn (changes requested):

sumitjnn (approved):

PR #215, opened 2023-03-24
Add DML logging at debug and trace level.

yjhawar:


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -623,8 +623,11 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
 
     latestOffset = event.getOffset();
     latestSequenceNum = sequenceNumber;
+
     if (LOG.isTraceEnabled()) {
-      LOG.trace("DML event:{} offset: {} seq num : {}", GSON.toJson(event), latestOffset.get(), latestSequenceNum);
+      LOG.trace("DML event={}, sequenceNumber={}", GSON.toJson(event), latestSequenceNum);
+    } else if (LOG.isDebugEnabled()) {
+      LOG.debug("DML Event={}, sequenceNumber={}", event, latestSequenceNum);
```
sumitjnn: Nit: let's log `sequenceNumber` instead of `latestSequenceNum`, although both are equivalent at this point, but semantically it makes it a little confusing, as we just want to log event's sequence number.

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -623,8 +623,11 @@ public synchronized void applyDML(Sequenced<DMLEvent> sequencedEvent) throws Exc
 
     latestOffset = event.getOffset();
     latestSequenceNum = sequenceNumber;
+
     if (LOG.isTraceEnabled()) {
-      LOG.trace("DML event:{} offset: {} seq num : {}", GSON.toJson(event), latestOffset.get(), latestSequenceNum);
+      LOG.trace("DML event={}, sequenceNumber={}", GSON.toJson(event), latestSequenceNum);
```
sumitjnn: nit: let's use `DML event` or `DML Event` consistently

src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
sumitjnn: It would be better to log the event after L#592, as if we will get the event details even if there are any exceptions in processing.

sumitjnn (approved):

PR #217, opened 2023-03-28
Logging original exception stacktrace by adding it as cause(CDAP-20476)

yjhawar:


sumitjnn:
@yjhawar Please add jira id to PR title

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -300,7 +300,7 @@ public static Job createBigQueryJob(BigQuery bigquery, JobInfo jobInfo) {
       return bigquery.create(jobInfo);
     } catch (BigQueryException e) {
       if (e.getCode() == 409 && BIG_QUERY_DUPLICATE_ERROR.equalsIgnoreCase(e.getReason())) {
-        LOG.warn("Got JOB ALREADY EXISTS for the job id : " + jobInfo.getJobId() + ". Returning existing job");
+        LOG.warn("Got JOB ALREADY EXISTS for the job id : " + jobInfo.getJobId() + ". Returning existing job", e);
```
sumitjnn: Exception is not needed here as it does not provide any additional value, we already know it's due to duplicate job so logging jobId should be enough.

sumitjnn (approved):

PR #236, opened 2023-08-08
Bump bigquery-delta-plugins version

Nikitapaliwal123:
Bump bigquery-delta-plugins version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
 
   <groupId>io.cdap.delta</groupId>
   <artifactId>bigquery-delta-plugins</artifactId>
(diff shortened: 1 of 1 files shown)


itsankit-google (approved):

itsankit-google (changes requested):
We cannot change the artifact version and remove snapshot, we need to bump the dependency version which was done in https://github.com/data-integrations/bigquery-delta-plugins/commit/88d797103584cdefd0a8496f6bd067df2b12d9d0#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R60

PR #246, opened 2023-10-23
[Plugin-1703] Support chinese column names

psainics:
## Support for Chinese character in column names [ Bug Fix ]

Jira : [Plugin-1703](https://cdap.atlassian.net/browse/PLUGIN-1703)

### Changes
- Enabled extended character support for field/column names.
- As per docs max column name can now be 300
- New regex for valid and invalid field/column names.
- Update function `isJsonFormat` to return true when non English character is used
   -  We do this as the avro load job in BQ does not support non-english characters in field names for now
   
### UI Field
- A new ui field is added (false by default) to maintain backwards compatibility, where users can trigger old mappings. 
image

### Unit Test 
- Ran old unit test as validation
image

- Extended test cases with
  - Contains hyphen 
  - Contains Chinese character 
  - Contains Japanese character
  - Contains Emoji

### DIFF
--- a/docs/bigquery-cdcTarget.md
+++ b/docs/bigquery-cdcTarget.md
@@ -82,6 +82,19 @@ name is same as source database name. A valid name should only contain letters, 
 maximum length can be 1024. Any invalid chars would be replaced with underscore in the final dataset name and
 any characters exceeds length limit will be truncated.
 
+**Allow Flexible Column Naming**:
+By default ony english letters, numbers and underscore are allowed in column names. If this option is enabled,
+international characters are also allowed in column names with some extra special characters, which follow the
+bigquery naming convention for flexible column names.
+Some special characters allowed in flexible column names are: 
+- An ampersand (&)
+- A percent sign (%)
+- A colon (:)
+- A lessthan sign (<)
+- A space ( )
+
+Read more about this option [here](https://cloud.google.com/bigquery/docs/schemas#flexible-column-names).
+
 **Encryption Key Name**: GCP Customer-managed encryption key (CMEK) used to encrypt the resources created by this target. 
 Encryption key name should be of the form "projects/<gcp-project-id>/locations/<key-location>/keyRings/<key-ring-name>/cryptoKeys/<key-name>".
 
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
@@ -57,7 +57,7 @@ import javax.annotation.Nullable;
  */
 public final class BigQueryUtils {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryUtils.class);
-  public static final int FIELD_NAME_MAX_LENGTH = 128;
+  public static final int FIELD_NAME_MAX_LENGTH = 300;
   static final String BACKTICK = "`";
   private static final int DATASET_OR_TABLE_NAME_MAX_LENGTH = 1024;
   // Valid BigQuery dataset names can contain only letters, numbers, and underscores.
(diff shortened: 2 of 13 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
```diff
@@ -68,6 +68,8 @@ public final class BigQueryUtils {
   // Pc (connector, including underscore), Pd (dash), Zs (space).
   // See here: https://cloud.google.com/bigquery/docs/tables#table_naming
   private static final Pattern VALID_TABLE_NAME_REGEX = Pattern.compile("[\\p{L}\\p{M}\\p{N}\\p{Pc}\\p{Pd}\\p{Zs}]+");
+  private static final Pattern VALID_FIELD_NAME_REGEX = Pattern.compile("[\\p{L}\\p{M}\\p{N}\\p{Pc}\\p{Pd}]+");
```
MrRahulSharma: Can you confirm if space is not allowed ? As per the documentation here : https://cloud.google.com/bigquery/docs/schemas#flexible-column-names, it conflicts.

psainics: Updated the PR with new requirements.

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
```diff
@@ -153,23 +153,34 @@ public void testNormalizeTableName() {
...
-      name = Strings.repeat("a1", 65);
-      assertEquals(name.substring(0, 128), BigQueryUtils.normalizeFieldName(name));
+      name = Strings.repeat("a1", 151);
+      assertEquals(name.substring(0, maxColumnNameLength), BigQueryUtils.normalizeFieldName(name, false));
       // contains invalid character
-      assertEquals("ab_c", BigQueryUtils.normalizeFieldName("ab?/c"));
+      assertEquals("ab_c", BigQueryUtils.normalizeFieldName("ab?/c", false));
       // contains space
-      assertEquals("a2_fs", BigQueryUtils.normalizeFieldName("a2 fs"));
+      assertEquals("a2_fs", BigQueryUtils.normalizeFieldName("a2 fs", false));
```
MrRahulSharma, line 175: Please add a test for whitespace comparison as well.

docs/bigquery-cdcTarget.md
```diff
@@ -82,6 +82,12 @@ name is same as source database name. A valid name should only contain letters,
 maximum length can be 1024. Any invalid chars would be replaced with underscore in the final dataset name and
 any characters exceeds length limit will be truncated.
 
+**Allow Flexible Column Naming**:
+By default ony english letters, numbers and underscore are allowed in column names. If this option is enabled,
+international characters are also allowed in column names with some extra special characters, which follow the
```
MrRahulSharma, line 87: It would help if we can list a few chars which will get supported by this change.

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
MrRahulSharma: What is the resolution here, are we supporting whitespaces in the field names ?

docs/bigquery-cdcTarget.md
psainics: Updated docs, incuded
> Some special characters allowed in flexible column names are: 
> - An ampersand (&)
> - A percent sign (%)
> - A colon (:)
> - A lessthan sign (<)
> - A space ( )

src/test/java/io/cdap/delta/bigquery/BigQueryUtilsTest.java
psainics: Great! I've added tests for whitespace and covered all the valid and invalid characters specified in the documentation.

src/main/java/io/cdap/delta/bigquery/BigQueryUtils.java
psainics: Yes, we are supporting all valid flexible names, with whitespaces and other special characters.
This is the new regex `"[\\p{L}\\p{M}\\p{N}\\p{Pc}\\p{Pd}&%+=:'<>#|\\s]+"`

docs/bigquery-cdcTarget.md
```diff
@@ -82,6 +82,19 @@ name is same as source database name. A valid name should only contain letters,
...
+bigquery naming convention for flexible column names.
+Some special characters allowed in flexible column names are: 
+- An ampersand (&)
+- A percent sign (%)
+- A colon (:)
+- A lessthan sign (<)
+- A space ( )
```
MrRahulSharma, line 94:
```suggestion
- Ampersand (&)
- Percent (%)
- Colon (:)
- Less than (<)
- Whitespace ( )
```

MrRahulSharma (approved):

PR #274, opened 2025-02-26
PLUGIN-1833: fix: Use unique aliases in MERGE statements to avoid conflicts with column names

anshumanks:
### Use unique aliases in MERGE statements to avoid conflicts with column names

Jira : [PLUGIN-1833](https://cdap.atlassian.net/browse/PLUGIN-1833)

### Description

This update resolves an issue where the MERGE operation was failing if the user's table contained columns A, B, or D. The failure occurred because these column names were being used as table aliases in the code, causing conflicts.

### Code change

- Modified `BigQueryEventConsumer.java`

### Tested
(Same configurations used in both cases)

Before changes:
```
2025-03-05 23:14:25,377 - ERROR [bq-daemon-3:i.c.d.b.BigQueryEventConsumer@1478] - Failed to merge a batch of changes from the staging table into test2.MyTable
com.google.cloud.bigquery.BigQueryException: Value has type STRUCT<_op STRING, _batch_id INT64, _sequence_num INT64, ...> which cannot be inserted into column D, which has type STRING at [14:71]
```

After changes:
```
2025-03-05 23:28:03,354 - INFO  [bq-daemon-0:i.c.d.b.BigQueryEventConsumer@796] - Loading batch 1741217199562 of 5 events into target table for test2.MyTable 
2025-03-05 23:29:33,352 - INFO  [bq-daemon-2:i.c.d.b.BigQueryEventConsumer@796] - Loading batch 1741217323821 of 1 events into staging table for test2.MyTable 
2025-03-05 23:29:36,950 - INFO  [bq-daemon-2:i.c.d.b.BigQueryEventConsumer@898] - Merging batch 1741217323821 for test2.MyTable
```

[PLUGIN-1833]: https://cdap.atlassian.net/browse/PLUGIN-1833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
+++ b/src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
@@ -1190,34 +1190,34 @@ public class BigQueryEventConsumer implements EventConsumer {
      *    This makes sure event in B happens later than event in A
      */
     if (sourceRowIdSupported) {
-      joinCondition = "A._row_id = B._row_id ";
-      whereClause = " B._row_id IS NULL ";
+      joinCondition = "__cdap_A__._row_id = __cdap_B__._row_id ";
+      whereClause = " __cdap_B__._row_id IS NULL ";
     } else {
       joinCondition = primaryKeys.stream()
-        .map(name -> String.format("A.`%s` = B.`_before_%s`", name, name))
+        .map(name -> String.format("__cdap_A__.`%s` = __cdap_B__.`_before_%s`", name, name))
         .collect(Collectors.joining(" AND "));
 
       whereClause = primaryKeys.stream()
-        .map(name -> String.format("B.`_before_%s` IS NULL", name))
+        .map(name -> String.format("__cdap_B__.`_before_%s` IS NULL", name))
         .collect(Collectors.joining(" AND "));
     }
 
     if (sourceEventsOrdering == SourceProperties.Ordering.ORDERED) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/delta/bigquery/BigQueryEventConsumer.java
```diff
@@ -1190,34 +1190,34 @@ static String createDiffQuery(TableId stagingTable, List<String> primaryKeys, lo
      *    This makes sure event in B happens later than event in A
      */
     if (sourceRowIdSupported) {
-      joinCondition = "A._row_id = B._row_id ";
-      whereClause = " B._row_id IS NULL ";
+      joinCondition = "__cdf_A__._row_id = __cdf_B__._row_id ";
```
itsankit-google: Please use cdap not cdf in opensource components.

sahusanket:
NIT : 
Can you please replace the alias Values in comments as well? 

There are multiple places where logic is explained using alias, so it will be easy to refer.

sahusanket (approved):

anshumanks:
> NIT : Can you please replace the alias Values in comments as well?
> 
> There are multiple places where logic is explained using alias, so it will be easy to refer.

I intentionally left it that way. When I changed the values in the comment, it became unreadable because of so many underscores.

